PR #1, opened 2021-05-25
Sync rapid pro to engagement db

as2388:
This PR includes what I think is the minimum viable implementation for syncing runs from Rapid Pro to an engagement database, as well as setting a proposed structure for what the new pipelines should look like. Compared to the existing pipelines, it organises all the tools into self-contained packages, with the top-level scripts being very thin drivers of those tools, but otherwise should feel familiar. The other major change is that it writes urns to the uuid table rather than phone numbers. This is to future-proof us against possible future work with other channels.

Test configuration is included that syncs avf employee test data from Experimental-Test to firebase-test. I've left the output from a run of this tool in firebase-test for you to check.

There's still lots to do, and I've tried to capture these as todos in the code. I think we have enough here so that building out drafts of the other tools is the next priority, but if there's anything I've missed or any todos which you think are risky and need to be worked through now, that would be very helpful to know.


src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -0,0 +1,155 @@
...
+    :param engagement_db: Engagement database to check for the message.
+    :type engagement_db: engagement_database.EngagementDatabase
+    :param message: Message to check for existence.
+    :type message: engagement_database.data_models.Message
+    :return: Whether a message with this text, timestamp, and participant_uuid exists in the engagement database.
+    :rtype: bool
+    """
+    # TODO: Rapid Pro has a bug where timestamps occasionally drift by 1us when runs are archived.
+    #       Confirm this is resolved before entering production with an '==' check on timestamps.
+    filter = lambda q: q \
```
IsaackMwenda, line 48: Checking for exists for single messages will be expensive as we scale. We should probably fetch message ids in cache and use this for the check.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -0,0 +1,155 @@
...
+        .where("text", "==", message.text) \
+        .where("timestamp", "==", message.timestamp) \
+        .where("participant_uuid", "==", message.participant_uuid)
```
IsaackMwenda, line 51: Let us also add the dataset name to narrow down the query scope.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -0,0 +1,155 @@
...
+    :param uuid_table: UUID table to use to de-identify contact urns.
+    :type uuid_table: id_infrastructure.firestore_uuid_table.FirestoreUuidTable
+    """
+    # This implementation is WIP. It shows how we can non-incrementally synchronise a workspace to the database.
+    # To enter production, we still need the following:
+    # TODO: Support incremental update of runs and contacts.
+    # TODO: Handle deleted contacts.
+    # TODO: Handle contacts that have runs but haven't been fetched locally yet.
+    # TODO: Optimise fetching fields from the same flows, so we don't have to download the same runs multiple times.
+
```
IsaackMwenda, line 105: TODO: Auto code before adding to database

sync_rapid_pro_to_engagement_db.py
```diff
@@ -0,0 +1,58 @@
...
+        table_name="_engagement_db_test",
+        uuid_prefix="avf-participant-uuid-"
+    )
+
+    engagement_db_configuration = EngagementDatabaseConfiguration(
+        credentials_file_url="gs://avf-credentials/firebase-test.json",
+        database_path="engagement_db_experiments/experimental_test"
+    )
+
+    rapid_pro_config = RapidProToEngagementDBConfiguration(
```
IsaackMwenda, line 43: We should probably move this to the config json file so that we can reuse the rest of the common configuration across pipelines

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -0,0 +1,155 @@
...
+                log.debug("No relevant run result")
+                continue
+
+            # De-identify the contact's urn.
+            # Note (1) this de-identifies the full urn, so we can handle multiple channel types simultaneously.
+            #      (2) we only get this far if there was a valid run result, which ensures we only add de-identification
+            #          entries for participants who messaged us.
+            contact_urn = contacts_lut[run.contact.uuid].urns[0]
+            if contact_urn.startswith("tel:"):
+                assert contact_urn.startswith("tel:+")
```
IsaackMwenda, line 135: This is not usually the case and not quite sure how we can resolve this because if we normalize while adding to db when we add normalised advert contacts back to textit, it  treats them as two different contacts. - probably build a case and reach out to them.

IsaackMwenda (approved):
This LGTM as a beta - have left a few comments and happy to discuss this further offline. Thanks for the good work.

sync_rapid_pro_to_engagement_db.py
as2388: Yes, will definitely do this in future PRs as the configuration stabilises.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: We can't add the dataset to the search because WS-correction means the dataset can change. As far as the sync from rapid pro is concerned, the dataset is just a best initial guess for how the rest of the pipeline should treat the message, but this can be changed later.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: I think this is best done when syncing to Coda, as per the design doc. Then all the logic and configuration related to labelling is all done in one place, the coda sync.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: I don't think this cost is very high, as in incremental mode we're only paying 1 read per incoming message by doing this.  How would a cache help?

DanielMwendwa:
I propose the following changes in configuration. I think it ensures a bit of consistency:

- Add RapidproConfiguration to common configurations and initialize rapid pro as uuid_table & engagement_db which means no RapidProToEngagementDBConfiguration. `rapid_pro_config = RapidProConfiguration(domain="", token_file_url="")`
- Initialize flow_result_configurations also like above ( uuid_table & engagement_db). `flow_result_configurations=[...]`
- Then `sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, flow_result_config)`

DanielMwendwa (approved):

DanielMwendwa:
Thanks @as2388.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
IsaackMwenda: I was thinking  - For the Initial rapid pro sync, we store message ids for a project in the cache, for the next subsequent run we check if the new message-id exists in the cached Ids and update the list write to firestore if it does not. This means we do a full read from a project firestore once at the pipeline initial run but read and write only when we have new messages.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: Following offline discussion this only impacts golis via shaqodoon, so I'll add a todo here but leave this fail-safe for now. We'll address fully once we apply this to a project that has this problem

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: resolved offline

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: I think we're in the same situation - If we add a project name (which can alternatively be encoded in the dataset e.g. CSAP_age), we create the possibility of adding the message multiple times if we change project.

as2388:
> I propose the following changes in configuration. I think it ensures a bit of consistency:
> 
> * Add RapidproConfiguration to common configurations and initialize rapid pro as uuid_table & engagement_db which means no RapidProToEngagementDBConfiguration. `rapid_pro_config = RapidProConfiguration(domain="", token_file_url="")`
> * Initialize flow_result_configurations also like above ( uuid_table & engagement_db). `flow_result_configurations=[...]`
> * Then `sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, flow_result_config)`

Thanks @DanielMwendwa, great suggestions. I will do this in the next PR so it's easy for you to review.

PR #2, opened 2021-05-27
Move RapidPro configuration to common/configuration

as2388:
This simplifies and improves the consistency in configuration, as well as making it easier to reuse rapid pro configuration in other tools in future

### DIFF
--- a/src/rapid_pro_to_engagement_db/configuration.py
+++ b/src/rapid_pro_to_engagement_db/configuration.py
@@ -6,10 +6,3 @@ class FlowResultConfiguration:
     flow_name: str
     flow_result_field: str
     engagement_db_dataset: str
-
-
-@dataclass
-class RapidProToEngagementDBConfiguration:
-    domain: str
-    token_file_url: str
-    flow_result_configurations: [FlowResultConfiguration]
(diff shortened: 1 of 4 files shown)


src/rapid_pro_to_engagement_db/configuration.py
```diff
@@ -6,10 +6,3 @@ class FlowResultConfiguration:
     flow_name: str
     flow_result_field: str
     engagement_db_dataset: str
```
DanielMwendwa, line 8: we should also move this to common/config

sync_rapid_pro_to_engagement_db.py
```diff
@@ -4,7 +4,7 @@
 from core_data_modules.logging import Logger
 from engagement_database.data_models import HistoryEntryOrigin
 
-from src.common.configuration import UUIDTableConfiguration, EngagementDatabaseConfiguration
+from src.common.configuration import UUIDTableConfiguration, EngagementDatabaseConfiguration, RapidProConfiguration
 from src.rapid_pro_to_engagement_db.configuration import RapidProToEngagementDBConfiguration, FlowResultConfiguration
```
DanielMwendwa, line 8: rm RapidProToEngagementDBConfiguration

DanielMwendwa (approved):

sync_rapid_pro_to_engagement_db.py
as2388: fixed, thanks

src/rapid_pro_to_engagement_db/configuration.py
as2388: I'm not sure we should - this configuration is specific to rapid pro -> db sync

DanielMwendwa: yeah sure. you can go ahead and merge.

PR #14, opened 2021-06-10
Enable incremental mode for rapid pro -> engagement db sync

as2388:
This works similarly to how the operations dashboard's sms service works - you can optionally pass a directory the script should use to cache results, so it only needs to download runs + contacts that changed since it last checked. Once dockerised, this would live in a docker volume and be reset simply by removing the volume.

### DIFF
--- a/src/rapid_pro_to_engagement_db/cache.py
+++ b/src/rapid_pro_to_engagement_db/cache.py
@@ -7,12 +7,28 @@ from temba_client.v2 import Contact
 
 class RapidProSyncCache(object):
     def __init__(self, cache_dir):
+        """
+        Initialises a Rapid Pro sync cache at the given directory.
+
+        The sync cache can be used to locally save/retrieve data needed to enable incremental running of a
+        Rapid Pro -> Engagement Database sync tool.
+
+        :param cache_dir: Directory to use for the cache.
+        :type cache_dir: str
+        """
+
         self.cache_dir = cache_dir
 
     def _contacts_path(self):
         return f"{self.cache_dir}/contacts.json"
 
     def get_contacts(self):
+        """
+        Gets cached contacts.
+
+        :return: Cached contacts, or None if there is no cache yet.
+        :rtype: list of temba_client.v2.Contact | None
+        """
         try:
             with open(self._contacts_path()) as f:
                 return [Contact.deserialize(d) for d in json.load(f)]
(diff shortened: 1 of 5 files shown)


IsaackMwenda (approved):

src/rapid_pro_to_engagement_db/cache.py
```diff
@@ -0,0 +1,36 @@
+from datetime import datetime
+import json
+
+from core_data_modules.util import IOUtils
+from temba_client.v2 import Contact
+
+
+class RapidProSyncCache(object):
```
DanielMwendwa, line 8: Since we're using python 3.x, do we need to inherit from object? Anyway I know most of our class inherit from object, if its for consistency I think you can just leave as is.

src/rapid_pro_to_engagement_db/cache.py
```diff
@@ -0,0 +1,36 @@
...
+            return None
+
+    def set_contacts(self, contacts):
+        IOUtils.ensure_dirs_exist_for_file(self._contacts_path())
+        with open(self._contacts_path(), "w") as f:
+            json.dump([c.serialize() for c in contacts], f)
+
+    def get_flow_last_updated(self, flow_id, result_field):
+        try:
+            with open(f"{self.cache_dir}/flow-{flow_id}-{result_field}-last-updated.txt") as f:
```
DanielMwendwa: I think we should have def  _last_updated_path(self, cache_dir, flow_id, result_field)

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -79,18 +81,42 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, flow_r
...
     for flow_config in flow_result_configs:
         # Get the latest runs for this flow.
         flow_id = rapid_pro.get_flow_id(flow_config.flow_name)
-        runs = rapid_pro.get_raw_runs(flow_id)
+        flow_last_updated = None
+        if cache is not None:
+            flow_last_updated = cache.get_flow_last_updated(flow_id, flow_config.flow_result_field)
+        fetch_since = None
+        if flow_last_updated is not None:
+            fetch_since = flow_last_updated + timedelta(microseconds=1)
```
DanielMwendwa: we can use timedelta.resolution, but feel free to leave as is.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -1,12 +1,58 @@
...
+    """
+    # Try to get the last modified timestamp from the cache
+    flow_last_updated = None
+    if cache is not None:
+        flow_last_updated = cache.get_latest_run_timestamp(flow_id, flow_result_field)
+
+    # If there is a last updated timestamp in the cache, only download and return runs that have been modified since.
+    filter_last_modified_after = None
+    if flow_last_updated is not None:
+        filter_last_modified_after = flow_last_updated + timedelta(microseconds=1)
```
DanielMwendwa, line 51: we can use timedelta.resolution, but feel free to leave as is.

DanielMwendwa (approved):

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: Thanks, I've checked the docs for .resolution and still think this is clearer because (i) we don't need to look up what the resolution is, (ii) this needs to be textit's resolution rather than python's, which are the same by coincidence.

src/rapid_pro_to_engagement_db/cache.py
as2388: No we don't, this was just habit. I will remove this here and try to stop using it in future, that way we need to understand one fewer of python's mistakes going forwards.

PR #16, opened 2021-06-11
Separate the continuous sync -> rapid pro configuration from initial sync configuration

as2388:
This separates what we'd expect to sync continuously as part of a pipeline from what we'd sync when creating a new workspace. For a new workspace, we sync all the relevant fields to make the workspace look like the engagement database, for a pipeline I think we only want to continuously sync consent status for now. Syncing anything else in a pipeline could lead to rapid pro re-asking the same question to the same person very often, which is not desirable.

The general idea here with multiple configuration files is that we can have a configuration file for every pipeline/manual operation we run on a project, so operations are very easy to code review and to run on different machines. Contrast this with some of our current manual tasks, that require a lot of command line arguments to be configured instead.

### DIFF
--- a/configurations/init_test_workspace.py
+++ b/configurations/init_test_workspace.py
@@ -1,0 +1,53 @@
+import json
+
+from core_data_modules.cleaners import swahili
+from core_data_modules.data_models import CodeScheme
+from temba_client.v2 import Field
+
+from src.common.configuration import RapidProClientConfiguration, CodaClientConfiguration, UUIDTableClientConfiguration, \
+    EngagementDatabaseClientConfiguration
+from src.engagement_db_to_coda.configuration import CodaSyncConfiguration, CodaDatasetConfiguration, \
+    CodeSchemeConfiguration
+from src.engagement_db_to_rapid_pro.configuration import EngagementDBToRapidProConfiguration, DatasetConfiguration, \
+    WriteModes, ContactField
+from src.pipeline_configuration_spec import PipelineConfiguration, RapidProSource, CodaConfiguration, RapidProTarget
+from src.rapid_pro_to_engagement_db.configuration import FlowResultConfiguration
+
+
+def load_code_scheme(fname):
+    with open(f"code_schemes/{fname}.json") as f:
+        return CodeScheme.from_firebase_map(json.load(f))
+
+
+PIPELINE_CONFIGURATION = PipelineConfiguration(
+    pipeline_name="init-test-workspace",
+    engagement_database=EngagementDatabaseClientConfiguration(
+        credentials_file_url="gs://avf-credentials/firebase-test.json",
+        database_path="engagement_db_experiments/experimental_test"
+    ),
+    uuid_table=UUIDTableClientConfiguration(
+        credentials_file_url="gs://avf-credentials/firebase-test.json",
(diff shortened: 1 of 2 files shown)


configurations/init_test_workspace.py
```diff
@@ -0,0 +1,53 @@
...
+from core_data_modules.cleaners import swahili
+from core_data_modules.data_models import CodeScheme
+from temba_client.v2 import Field
+
+from src.common.configuration import RapidProClientConfiguration, CodaClientConfiguration, UUIDTableClientConfiguration, \
+    EngagementDatabaseClientConfiguration
+from src.engagement_db_to_coda.configuration import CodaSyncConfiguration, CodaDatasetConfiguration, \
+    CodeSchemeConfiguration
+from src.engagement_db_to_rapid_pro.configuration import EngagementDBToRapidProConfiguration, DatasetConfiguration, \
+    WriteModes, ContactField
```
DanielMwendwa, line 12: consider using tuples to do multi-line imports
Screenshot 2021-06-14 at 15 12 36

DanielMwendwa (approved):

DanielMwendwa:
> This separates what we'd expect to sync continuously as part of a pipeline from what we'd sync when creating a new workspace. For a new workspace, we sync all the relevant fields to make the workspace look like the engagement database, for a pipeline I think we only want to continuously sync consent status for now. Syncing anything else in a pipeline could lead to rapid pro re-asking the same question to the same person very often, which is not desirable.
> 
> The general idea here with multiple configuration files is that we can have a configuration file for every pipeline/manual operation we run on a project, so operations are very easy to code review and to run on different machines. Contrast this with some of our current manual tasks, that require a lot of command line arguments to be configured instead.

Noted, thanks.

configurations/init_test_workspace.py
```diff
@@ -0,0 +1,53 @@
+import json
+
+from core_data_modules.cleaners import swahili
+from core_data_modules.data_models import CodeScheme
+from temba_client.v2 import Field
+
+from src.common.configuration import RapidProClientConfiguration, CodaClientConfiguration, UUIDTableClientConfiguration, \
+    EngagementDatabaseClientConfiguration
+from src.engagement_db_to_coda.configuration import CodaSyncConfiguration, CodaDatasetConfiguration, \
```
IsaackMwenda, line 9: We should clean up unused imports here and add them when needed.

configurations/init_test_workspace.py
```diff
@@ -0,0 +1,53 @@
...
+    CodeSchemeConfiguration
+from src.engagement_db_to_rapid_pro.configuration import EngagementDBToRapidProConfiguration, DatasetConfiguration, \
+    WriteModes, ContactField
+from src.pipeline_configuration_spec import PipelineConfiguration, RapidProSource, CodaConfiguration, RapidProTarget
+from src.rapid_pro_to_engagement_db.configuration import FlowResultConfiguration
+
+
+def load_code_scheme(fname):
```
IsaackMwenda, line 17: ditto

IsaackMwenda (approved):

configurations/init_test_workspace.py
as2388: Will address this shortly in a new pr

PR #20, opened 2021-06-16
Update engagement db -> Coda sync to fetch and write message by message

as2388:
Previously we fetched all the new documents, then updated them message by message. This removes the double-fetch read cost from that set-up, as well as laying the foundations for this to easily switch to incremental running. It will also now be really easy to batch these updates in future, simply by increasing the .limit()s and iterating over the fetched messages in _sync_next_message_to_coda

### DIFF
--- a/src/engagement_db_to_coda/engagement_db_to_coda.py
+++ b/src/engagement_db_to_coda/engagement_db_to_coda.py
@@ -1,4 +1,3 @@
-import core_data_modules.data_models
 from core_data_modules.cleaners.cleaning_utils import CleaningUtils
 from core_data_modules.data_models import Message as CodaMessage
 from core_data_modules.logging import Logger
--- a/src/engagement_db_to_coda/engagement_db_to_coda.py
+++ b/src/engagement_db_to_coda/engagement_db_to_coda.py
@@ -1,11 +1,9 @@
-import json
-
 import core_data_modules.data_models
 from core_data_modules.cleaners.cleaning_utils import CleaningUtils
+from core_data_modules.data_models import Message as CodaMessage
 from core_data_modules.logging import Logger
 from core_data_modules.util import SHAUtils, TimeUtils
 from engagement_database.data_models import HistoryEntryOrigin, MessageStatuses
-from core_data_modules.data_models import Message as CodaMessage
 from google.cloud import firestore
 
 log = Logger(__name__)
@@ -140,11 +138,12 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
 
 
 @firestore.transactional
-def _sync_message_to_coda(transaction, engagement_db, coda, coda_config, message_id):
+def _sync_next_message_to_coda(transaction, engagement_db, coda, coda_config, last_seen_message):
     """
     Syncs a message from an engagement database to Coda.
 
     This method:
+     - Gets the least recently updated message that was last updated after `last_updated_since`.
      - Writes back a coda id if the engagement db message doesn't have one yet.
      - Syncs the labels from Coda to this message if the message already exists in Coda.
      - Creates a new message in Coda if this message hasn't been seen in Coda yet.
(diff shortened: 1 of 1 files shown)


src/engagement_db_to_coda/engagement_db_to_coda.py
```diff
@@ -155,10 +154,40 @@ def _sync_message_to_coda(transaction, engagement_db, coda, coda_config, message
...
+                              downloads the least recently updated message in the database.
+    :type last_seen_message: engagement_database.data_models.Message | None
+    :return: The engagement database message that was synced.
+             If there was no new message to sync, returns None.
+    :rtype: engagement_database.data_models.Message | None
     """
-    engagement_db_message = engagement_db.get_message(message_id, transaction=transaction)
+    if last_seen_message is None:
+        messages_filter = lambda q: q \
+            .where("status", "in", [MessageStatuses.LIVE, MessageStatuses.STALE]) \
```
IsaackMwenda, line 169: do we expect a message without any of these two status?

IsaackMwenda (approved):
lgtm thanks.

src/engagement_db_to_coda/engagement_db_to_coda.py
as2388: yes, the other 2 statuses we discussed were archived and anonymised, and we don't want to analyse messages with these statuses so no need to sync to coda

src/engagement_db_to_coda/engagement_db_to_coda.py
```diff
@@ -1,11 +1,9 @@
-import json
-
 import core_data_modules.data_models
```
DanielMwendwa, line 1: rm this import

src/engagement_db_to_coda/engagement_db_to_coda.py
```diff
@@ -155,10 +154,40 @@ def _sync_message_to_coda(transaction, engagement_db, coda, coda_config, message
...
+        # Get the next message modified at or later than the `last_seen_message`, excluding the `last_seen_message`.
+        messages_filter = lambda q: q \
+            .where("status", "in", [MessageStatuses.LIVE, MessageStatuses.STALE]) \
+            .order_by("last_updated") \
+            .order_by("message_id") \
+            .where("last_updated", ">=", last_seen_message.last_updated) \
+            .start_after({"last_updated": last_seen_message.last_updated, "message_id": last_seen_message.message_id}) \
+            .limit(1)
+
+    next_message_results = engagement_db.get_messages(filter=messages_filter, transaction=transaction)
```
DanielMwendwa, line 183: I think we should change param filter to firestore_filter or sth similar, to differentiate from the inbuilt module.

src/engagement_db_to_coda/engagement_db_to_coda.py
```diff
@@ -140,11 +138,12 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
 
 
 @firestore.transactional
-def _sync_message_to_coda(transaction, engagement_db, coda, coda_config, message_id):
+def _sync_next_message_to_coda(transaction, engagement_db, coda, coda_config, last_seen_message):
     """
     Syncs a message from an engagement database to Coda.
 
     This method:
+     - Gets the least recently updated message that was last updated after `last_updated_since`.
```
DanielMwendwa, line 146: There isn't variable name `last_updated_since` although it makes sense. You can say  ...after the last update or sth similar

DanielMwendwa (approved):

src/engagement_db_to_coda/engagement_db_to_coda.py
as2388: thanks, removed

src/engagement_db_to_coda/engagement_db_to_coda.py
as2388: great catch, this docstring is out of date so I've changed this to last_seen_message

src/engagement_db_to_coda/engagement_db_to_coda.py
as2388: have added as a low priority todo

PR #22, opened 2021-06-18
Sync to Coda dataset by dataset

as2388:
As previously discussed, this reduces the number of messages we need to fetch and means we don't need to clear any caches when adding new datasets, further reducing costs.

### DIFF
--- a/src/engagement_db_to_coda/engagement_db_to_coda.py
+++ b/src/engagement_db_to_coda/engagement_db_to_coda.py
@@ -138,7 +138,7 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
 
 
 @firestore.transactional
-def _sync_next_message_to_coda(transaction, engagement_db, coda, coda_config, last_seen_message):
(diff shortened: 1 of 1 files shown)


src/engagement_db_to_coda/engagement_db_to_coda.py
```diff
@@ -138,7 +138,7 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
 
 
 @firestore.transactional
-def _sync_next_message_to_coda(transaction, engagement_db, coda, coda_config, last_seen_message):
+def _sync_next_message_to_coda(transaction, engagement_db, coda, coda_config, dataset_config, last_seen_message):
```
DanielMwendwa, line 141: update docstring

DanielMwendwa (approved):

IsaackMwenda (approved):

src/engagement_db_to_coda/engagement_db_to_coda.py
as2388: thanks for catching this

PR #23, opened 2021-06-18
Add configuration for creating the Kenya Pool

as2388:
Creates the pool from all the demographic responses from projects where we asked people if they wanted to opt-in to the pool. (The alternative would have been to import from traced data - but this requires building and testing a separate tool, handling 'synthetic messages' in the traced data's demographics, and dealing with different history for those messages vs. all other messages).

This needs to be modified to only copy contacts who opted-in rather than everyone, but opening now so you have this while I'm on leave.

I've also added an optional description field to the configuration here as I think it will be useful to be able to briefly describe what a configuration does.

### DIFF
--- a/configurations/create_kenya_pool.py
+++ b/configurations/create_kenya_pool.py
@@ -1,0 +1,156 @@
+import json
+
+from core_data_modules.cleaners import swahili
+from core_data_modules.data_models import CodeScheme
+
+from src.common.configuration import RapidProClientConfiguration, CodaClientConfiguration, UUIDTableClientConfiguration, \
+    EngagementDatabaseClientConfiguration
+from src.engagement_db_to_coda.configuration import CodaSyncConfiguration, CodaDatasetConfiguration, \
+    CodeSchemeConfiguration
+from src.engagement_db_to_rapid_pro.configuration import EngagementDBToRapidProConfiguration, DatasetConfiguration, \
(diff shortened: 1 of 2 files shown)


configurations/create_kenya_pool.py
```diff
@@ -0,0 +1,156 @@
...
+
+def load_code_scheme(fname):
+    with open(f"code_schemes/{fname}.json") as f:
+        return CodeScheme.from_firebase_map(json.load(f))
+
+
+PIPELINE_CONFIGURATION = PipelineConfiguration(
+    pipeline_name="Create-Kenya-Pool",
+    description="Creates the initial Kenya Pool from demographics responses to COVID19, COVID19-Ke-Urban, "
+                "UNDP-Kenya, UNICEF-KENYA, OXFAM-KENYA, and World Vision.",
```
IsaackMwenda, line 24:
```suggestion
                "UNDP-Kenya, UNICEF-KENYA, OXFAM-KENYA, World Vision and GPSDD",
```

IsaackMwenda (approved):
lgtm

DanielMwendwa (approved):

configurations/create_kenya_pool.py
as2388: converted to a todo to add in a future pr

PR #26, opened 2021-06-29
Filter messages

IsaackMwenda:
This adds support to;
1. Convert messages to traced data objects.
2. Filter a list of td for research question messages received within the given time range.
3. Filter messages tagged as test_run. Have updated Pipeline Infrastructure https://github.com/AfricasVoices/Pipeline-Infrastructure/pull/56 to include test run property in Message object and updated rapid pro->db to check and sync this.


src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -194,6 +198,8 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, flow_r
             contact_urn = contact.urns[0]
             participant_uuid = _de_identify_contact_urn(contact_urn, uuid_table)
 
+            test_run = run.contact.uuid in test_contacts
```
as2388, line 201: I'm not sure about this strategy because it means we have to have the test contact uuids correct when we first copy the message to the database, and if we change who we consider to be a test contact (e.g. because someone at avf tested the line without checking with us first) that could be quite expensive.

Perhaps it would make more sense to change the test contacts uuid configuration to refer to participant uuids rather than rapid pro contact uuids. Then it's super easy to change who we consider test contacts, and we can more easily copy these across projects where these ids don't change (unlike rapid pro, which assigns a new uuid in every workspace).

configurations/test_pipeline_configuration.py
```diff
@@ -1,9 +1,18 @@
...
 from src.pipeline_configuration_spec import *
 
 PIPELINE_CONFIGURATION = PipelineConfiguration(
     pipeline_name="engagement-db-test",
+    project_start_date = isoparse("2021-03-01T10:30:00+03:00"),
+    project_end_date = isoparse("2100-01-01T00:00:00+03:00"),
+    filter_test_messages = False,
+    test_contact_uuids = [
+        "f9158369-b985-4271-a9d6-5dd25e19e43c"
+    ],
```
as2388, line 15: Do we want to apply these filters everywhere or just in analysis?

src/engagement_db_to_analysis/configuration.py
```diff
@@ -0,0 +1,13 @@
...
+
+@dataclass()
+class DatasetTypes:
+    DEMOGRAPHIC = "demographic"
+    RESEARCH_QUESTION_ANSWER = "research_question_answer"
+
+
+@dataclass
+class AnalysisConfiguration:
+    engagement_db_dataset: str
```
as2388, line 12: (If this is a str[] we get a lot more flexibility e.g. we can combine locations (as was needed by gpsdd)

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -35,7 +37,7 @@ def get_project_messages_from_engagement_db(dataset_configurations, engagement_d
     log.info(f"Initialising EngagementAnalysisCache at'{cache_path}'")
```
as2388, line 37:
```suggestion
    log.info(f"Initialising EngagementAnalysisCache at '{cache_path}'")
```

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -89,10 +93,48 @@ def get_project_messages_from_engagement_db(dataset_configurations, engagement_d
 
     return engagement_db_dataset_messages_map
 
-#TODO: Filter Messages
-def filter_messages():
-    return None
+def _convert_messages_to_traced_data(user, messages_map):
+    #TODO: add docsting
+
+    data = []
```
as2388, line 99: Can we call this what it is e.g. messages, or messages_td?

src/engagement_db_to_analysis/message_filters.py
```diff
@@ -0,0 +1,79 @@
...
+        # Exclusive end time of the time range to keep. Messages sent after this time will be dropped.
+        end_time_inclusive = pipeline_config.project_end_date
+
+        log.debug(f"Filtering out research question messages sent outside the project time range "
+                  f"{start_time_inclusive.isoformat()} to {end_time_inclusive.isoformat()}...")
+
+        # Filter a list of td for research question messages received within the given time range.
+        rqa_engagement_db_datasets = []
+        for analysis_config in pipeline_config.analysis_config:
+            if analysis_config.dataset_type == "research_question_answer":
```
as2388, line 63:
```suggestion
            if analysis_config.dataset_type == DatasetTypes.RESEARCH_QUESTION_ANSWER
```

src/engagement_db_to_analysis/configuration.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+@dataclass()
+class DatasetTypes:
+    DEMOGRAPHIC = "demographic"
+    RESEARCH_QUESTION_ANSWER = "research_question_answer"
+
+
+@dataclass
+class AnalysisConfiguration:
```
as2388, line 11: I think this is an AnalysisDatasetConfiguration rather than an AnalysisConfiguration, as it configures one dataset rather than a full analysis.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
IsaackMwenda: Thanks for noting this. I see how filtering using participant uids and in analysis is a better approach than in rapidpro fetch as it used to be. Have removed the filters, for now, will filter test participants after merging the messages traced data in individuals traced data.

configurations/test_pipeline_configuration.py
IsaackMwenda: At the moment yes - will keep this here and add them to the analysis configuration if no other module uses them.

src/engagement_db_to_analysis/configuration.py
IsaackMwenda: Quite right. Updated to list of strings.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
IsaackMwenda: updated thanks!

src/engagement_db_to_analysis/message_filters.py
IsaackMwenda: updated thanks!

src/engagement_db_to_analysis/configuration.py
IsaackMwenda: updated thanks!

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -2,28 +2,30 @@
...
 
 log = Logger(__name__)
 
 
 def serialise_message(msg):
-
     msg = msg.to_dict()
-    msg["timestamp"] = TimeUtils.datetime_to_utc_iso_string(msg["timestamp"])
-    msg["last_updated"] = TimeUtils.datetime_to_utc_iso_string(msg["last_updated"])
+    msg["timestamp"] = msg["timestamp"].isoformat()
```
as2388, line 15: I think we should Message.add to_json and .from_json to pipeline-infrastructure, maybe add a todo here if you agree

src/engagement_db_to_analysis/message_filters.py
```diff
@@ -0,0 +1,67 @@
...
+from core_data_modules.traced_data import Metadata
+
+from src.pipeline_configuration_spec import *
+
+
+log = Logger(__name__)
+
+
+# TODO: Move to Core.
+class Filters(object):
```
as2388, line 14: Given that these are all static functions, the more pythonic way to write this is to put the functions in a module, rather than wrapping in a class like this (I too was wrong about this for a long time, as in other languages this approach would be normal).

src/pipeline_configuration_spec.py
```diff
@@ -38,9 +39,13 @@ class RapidProTarget:
 @dataclass
 class PipelineConfiguration:
     pipeline_name: str
+    project_start_date: str
```
as2388, line 42: We talked offline about the possibility of tidying this up into e.g. filters = [TimeFilter(...), UuidFilter(...)] etc. and structuring configuration so that we don't need to pass the entire configuration into each individual step. But I think these are not priorities so fine to add a todo for those for now

as2388 (approved):

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -2,28 +2,30 @@
...
+    msg["last_updated"] = msg["last_updated"].isoformat()
 
     return msg
 
-def get_project_messages_from_engagement_db(dataset_configurations, engagement_db, cache_path=None):
+
+def _get_project_messages_from_engagement_db(analysis_configurations, engagement_db, cache_path=None):
```
DanielMwendwa, line 21: It is not clear that `cache_path` set to None while in the main `engagement_db_to_analysis.py` `engagement_db_datasets_cache_dir` is a required argument.

DanielMwendwa (approved):

src/engagement_db_to_analysis/engagement_db_to_analysis.py
IsaackMwenda: Yes now that it's fully tested in two areas we should move it. Added a Todo.

src/pipeline_configuration_spec.py
IsaackMwenda: Added a todo.

src/engagement_db_to_analysis/message_filters.py
IsaackMwenda: refactored to `message_filters` module.

src/engagement_db_to_analysis/configuration.py
```diff
@@ -0,0 +1,13 @@
+from dataclasses import dataclass
+
+
+@dataclass()
```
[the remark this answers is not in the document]
IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/message_filters.py
```diff
@@ -0,0 +1,82 @@
...
+from core_data_modules.traced_data import Metadata
+
+from src.pipeline_configuration_spec import *
+
+
+log = Logger(__name__)
+
+
+# TODO: Move to Core.
+class MessageFilters(object):
```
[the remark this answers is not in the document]
IsaackMwenda: refactored  to a module.

src/engagement_db_to_analysis/__init__.py
```diff
@@ -0,0 +1 @@
+from .message_filters import Filters
```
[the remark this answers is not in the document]
IsaackMwenda: Thanks, I have renamed and refactored it to `traced_data_filters.py` module.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -32,67 +34,99 @@ def get_project_messages_from_engagement_db(dataset_configurations, engagement_d
...
 
-        engagement_db_dataset_messages_map[engagement_db_dataset] = messages
+            # Update latest_message_timestamp
+            for msg in messages:
+                if latest_message_timestamp is None or isoparse(msg["last_updated"]) > latest_message_timestamp:
+                    latest_message_timestamp = isoparse(msg["last_updated"])
 
-        # Update latest_message_timestamp
-        for msg in messages:
-            if latest_message_timestamp is None or isoparse(msg["last_updated"]) > latest_message_timestamp:
```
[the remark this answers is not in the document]
IsaackMwenda: Refactored to
```
            for msg in messages:
                msg_last_updated = isoparse(msg["last_updated"])
                if latest_message_timestamp is None or msg_last_updated > latest_message_timestamp:
                    latest_message_timestamp = msg_last_updated
```
Thanks.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -32,67 +34,99 @@ def get_project_messages_from_engagement_db(dataset_configurations, engagement_d
...
+    :param messages_map: Dict containing messages data.
+    :type messages_map: dict
+    :return: A list of Traced data message objects.
+    :type: list of Traced data
+    """
+
+    messages_td = []
+    for engagement_db_dataset in messages_map:
+
+        engagement_db_dataset_messages = messages_map[engagement_db_dataset]
```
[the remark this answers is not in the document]
IsaackMwenda: The step-by-step breakdown makes this more readable and easier to follow. Leaving this as is.

src/engagement_db_to_analysis/message_filters.py
```diff
@@ -0,0 +1,67 @@
+from core_data_modules.logging import Logger
+from dateutil.parser import isoparse
+import time
+
+from core_data_modules.traced_data import Metadata
```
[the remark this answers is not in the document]
IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/message_filters.py
```diff
@@ -0,0 +1,67 @@
...
+
+        log.info(f"Filtered out messages sent outside the time range "
+                 f"{start_time_inclusive.isoformat()} to {end_time_inclusive.isoformat()}. "
+                 f"Returning {len(filtered)}/{len(data)} messages.")
+
+        return filtered
+
+
+    @classmethod
+    def filter_messages(cls, user, data, pipeline_config):
```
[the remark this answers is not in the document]
IsaackMwenda: renamed to messages_traced_data

src/engagement_db_to_analysis/message_filters.py
```diff
@@ -0,0 +1,67 @@
...
+        :type PIPELINE_CONFIGURATION:
+        :return: Filtered list.
+        :rtype: list of TracedData
+        """
+
+        # Inclusive start time of the time range to keep. Messages sent before this time will be dropped.
+        start_time_inclusive = pipeline_config.project_start_date
+
+        # Exclusive end time of the time range to keep. Messages sent after this time will be dropped.
+        end_time_inclusive = pipeline_config.project_end_date
```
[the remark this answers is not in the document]
IsaackMwenda: I strongly believe we should keep this. It may seem obvious to us, but super helpful documentation for new engineers.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
IsaackMwenda: Thanks for catching it, it was forgotten here after a previous refactor. Removed thanks.

PR #30, opened 2021-07-06
Fold messages by uid

IsaackMwenda:
Review https://github.com/AfricasVoices/Test-Pipeline-Engagement-DB/pull/26 first

PR #31, opened 2021-07-06
Filter test uids

IsaackMwenda:
Review https://github.com/AfricasVoices/Test-Pipeline-Engagement-DB/pull/30 and https://github.com/AfricasVoices/Test-Pipeline-Engagement-DB/pull/26 first


IsaackMwenda:
Updated and ready for review @as2388 @DanielMwendwa.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -168,4 +169,6 @@ def generate_analysis_files(user, pipeline_config, engagement_db, engagement_db_
 
     participants_traced_data_map = _fold_messages_by_uid(user, messages_traced_data)
```
as2388, line 170: Sorry I missed this before - does it make sense to pass this around as a map or as a list instead? I can see that on L172 you treat it like a list

src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -60,15 +60,50 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
...
+    :rtype: dict of participant_uid -> TracedData
+    """
+    log.debug("Filtering out test participants data...")
+    filtered = {}
+
+    for uid, trace_data in participants_traced_data_map.items():
+        if uid in test_contacts:
+            continue
+
+        trace_data.append_data(trace_data, Metadata(user, Metadata.get_call_location(), time.time()))
```
as2388, line 86:
```suggestion
        trace_data.append_data(trace_data, Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string()))
```
time.time() was used before we understood how to safely create timestamps in a way which was human readable too

src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -60,15 +60,50 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
...
+    :type participants_traced_data_map: dict of TracedData
+    :param test_contacts: a list containing test participant uids.
+    :type test_contacts: list of str
+    :return: Filtered dict.
+    :rtype: dict of participant_uid -> TracedData
+    """
+    log.debug("Filtering out test participants data...")
+    filtered = {}
+
+    for uid, trace_data in participants_traced_data_map.items():
```
as2388, line 82: I think we should always say what the traced_data is e.g. use participant here (or possibly participant_traced_data), otherwise you always have to think about whether this refers to one traceddata history or to a list, and whether its messages or participants

src/pipeline_configuration_spec.py
```diff
@@ -40,11 +40,12 @@ class RapidProTarget:
 @dataclass
 class PipelineConfiguration:
     pipeline_name: str
+    test_participant_uids: []
```
as2388, line 43: why do we need both test participant uids and filter_test_participants? Can we just have test_participant_uids: [] | None and only filter when the ids are specified?

as2388, line 43:
```suggestion
    test_participant_uuids: []
```
for consistency with the db

src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -60,15 +60,50 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
...
 def filter_messages(user, messages_data, pipeline_config):
 
     # Filter out runs sent outwith the project start and end dates
     messages_data = rqa_time_range_filter(user, messages_data, pipeline_config)
 
     return messages_data
+
+
+def filter_participants(user, participants_traced_data_map, pipeline_config):
+
```
as2388, line 102:
```suggestion
```

configurations/test_pipeline_configuration.py
```diff
@@ -8,9 +8,12 @@
...
     pipeline_name="engagement-db-test",
     # TODO: store in messages and individuals_filter list of functions.
-    project_start_date=isoparse("2021-03-01T10:30:00+03:00"),
-    project_end_date=isoparse("2100-01-01T00:00:00+03:00"),
-    filter_test_messages=False,
+    project_start_date = isoparse("2021-03-01T10:30:00+03:00"),
+    project_end_date = isoparse("2100-01-01T00:00:00+03:00"),
+    filter_test_participants = False,
+    test_participant_uids = [
+        "avf-participant-uuid-51c15546-58a0-4ab1-b465-e65b71462a8f"
```
as2388, line 15:
```suggestion
    project_start_date=isoparse("2021-03-01T10:30:00+03:00"),
    project_end_date=isoparse("2100-01-01T00:00:00+03:00"),
    filter_test_participants=False,
    test_participant_uids=[
        "avf-participant-uuid-51c15546-58a0-4ab1-b465-e65b71462a8f"
```
for consistency with the rest of our this file, the rest of avf's code base, and the official python formatting standards

as2388 (approved):

DanielMwendwa (approved):

configurations/test_pipeline_configuration.py
IsaackMwenda: Updated for consistency with the rest of this file - (If am not wrong pep8 recommends a single space on either side of the operator)

src/engagement_db_to_analysis/engagement_db_to_analysis.py
IsaackMwenda: L172 treats its as map. It was previously a list

src/engagement_db_to_analysis/traced_data_filters.py
IsaackMwenda: updated thanks

src/engagement_db_to_analysis/traced_data_filters.py
IsaackMwenda: updated to participant_trace_data

src/pipeline_configuration_spec.py
IsaackMwenda: updated thanks.

IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
as2388: sorry - I mean it treats it like a list in the sense that the operation doesn't take advantage of any of the features of a map, and could be trivially rewritten to use a list, as you could easily say
```
 for participant_traced_data in participants_traced_data:
         if participant_traced_data["uid"] in test_participant_uuids:
```

PR #33, opened 2021-07-08
Import modules as tuple to group the imports

DanielMwendwa:


as2388 (approved):
Thanks, this looks like the standard but not the pycharm default :/

(fix with preferences -> editor -> code style -> python -> wrapping and braces -> from import statements -> force parentheses if newline)

IsaackMwenda (approved):

PR #35, opened 2021-07-08
Add incremental mode to the coda -> engagement db sync

as2388:
Note this fetches and processes all changed messages, which makes it more expensive under failures than the engagement db -> coda sync (which syncs one-by-one). However, solving this is very difficult while Coda has segments so suggest we live with this for now.

### DIFF
--- a/sync_coda_to_engagement_db.py
+++ b/sync_coda_to_engagement_db.py
@@ -12,6 +12,8 @@ log = Logger(__name__)
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Syncs data from Coda to an engagement database")
 
+    parser.add_argument("--incremental-cache-path",
+                        help="Path to a directory to use to cache results needed for incremental operation.")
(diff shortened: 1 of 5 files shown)


DanielMwendwa (approved):

sync_coda_to_engagement_db.py
```diff
@@ -12,6 +12,8 @@
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Syncs data from Coda to an engagement database")
 
+    parser.add_argument("--incremental-cache-path",
```
IsaackMwenda, line 15: What's the benefit of making this optional?

IsaackMwenda (approved):

sync_coda_to_engagement_db.py
as2388: Primarily in development - it's easier to think about caches only when you need them rather than constantly needing to create and specify folders for them, remember what's in them, and manually delete them before every run.

PR #37, opened 2021-07-08
Fix type error in datetime filters

as2388:


configurations/test_pipeline_configuration.py
```diff
@@ -8,9 +8,9 @@
 PIPELINE_CONFIGURATION = PipelineConfiguration(
     pipeline_name="engagement-db-test",
     # TODO: store in messages and individuals_filter list of functions.
-    project_start_date = isoparse("2021-03-01T10:30:00+03:00"),
-    project_end_date = isoparse("2100-01-01T00:00:00+03:00"),
-    filter_test_messages = False,
+    project_start_date=isoparse("2021-03-01T10:30:00+03:00"),
```
IsaackMwenda, line 11:
```suggestion
    project_start_date = isoparse("2021-03-01T10:30:00+03:00"),
```
I think we need whitespace around the operator.

IsaackMwenda (approved):

configurations/test_pipeline_configuration.py
as2388: This would go against convention https://www.python.org/dev/peps/pep-0008/#other-recommendations

PR #43, opened 2021-07-14
Generates production file

DanielMwendwa:


src/engagement_db_to_analysis/generate_production_file.py
```diff
@@ -0,0 +1,7 @@
+from core_data_modules.traced_data.io import TracedDataCSVIO
+
+
+def generate_production_file(messages_td, production_csv_output_path):
+    production_keys = ["participant_uuid", "dataset", "text"]
+    with open(production_csv_output_path, "w") as f:
+        TracedDataCSVIO.export_traced_data_iterable_to_csv(messages_td, f, production_keys)
```
as2388, line 7: I don't think this is quite right as it just outputs a list of messages. The production file is formatted to give people easy access to quotes they can use on the radio, e.g. "a woman from nairobi said...", which means it needs to list demographics next to each message

engagement_db_to_analysis.py
```diff
@@ -21,18 +21,21 @@
     parser.add_argument("configuration_module",
                         help="Configuration module to import e.g. 'configurations.test_config'. "
                              "This module must contain a PIPELINE_CONFIGURATION property")
+    parser.add_argument("production_csv_output_path", metavar="production-csv-output-path",
```
as2388, line 24: can this be an analysis_dir instead? I think allowing this to be specified separately is unnecessary and makes setting up pipelines slower.

as2388 (changes requested):

src/engagement_db_to_analysis/generate_production_file.py
DanielMwendwa: Thanks, I'm working on it.

DanielMwendwa:
Made some progress, There are some issues i.e.,  the rows are duplicated and as you'll see there's a variable `messages` which is nested and I have to access it as `messages[0]`, this is because of `participants_traced_data_map` format. Also one cannot see participants who have only responded to the demographic quiz only, I'm not sure if this is the way production files are supposed to be.

@as2388 @IsaackMwenda, I've put up these changes so that you can review them as I am addressing the issues above and also get to know what your suggestions are concerning the issues above. Thanks yall.

DanielMwendwa:
Continues in #70

PR #48, opened 2021-07-19
Check that ws_codes are labelled as both WS in a normal scheme and as a code in the WS scheme

as2388:
This can be extended in future to insert CE codes here if we wish

### DIFF
--- a/src/engagement_db_coda_sync/lib.py
+++ b/src/engagement_db_coda_sync/lib.py
@@ -1,3 +1,4 @@
+from core_data_modules.cleaners import Codes
 from core_data_modules.cleaners.cleaning_utils import CleaningUtils
 from core_data_modules.data_models import Message as CodaMessage
 from core_data_modules.logging import Logger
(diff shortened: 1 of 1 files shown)


src/engagement_db_coda_sync/lib.py
```diff
@@ -67,6 +68,72 @@ def _add_message_to_coda(coda, coda_dataset_config, ws_correct_dataset_code_sche
...
+    ws_code = None
+    for label in coda_message.get_latest_labels():
+        if label.scheme_id == ws_code_scheme.scheme_id:
+            code_in_ws_scheme = True
+            ws_code = ws_code_scheme.get_code_with_code_id(label.code_id)
+
+    # Ensure there is a WS code in a normal scheme and a code in the WS scheme.
+    # If there isn't, don't attempt any redirect, so we can impute a CE code later.
+    if ws_code_in_normal_scheme != code_in_ws_scheme:
+        # TODO: Impute CE here?
```
IsaackMwenda, line 129: sgtm!

IsaackMwenda (approved):

src/engagement_db_coda_sync/lib.py
```diff
@@ -67,6 +68,72 @@ def _add_message_to_coda(coda, coda_dataset_config, ws_correct_dataset_code_sche
...
+        if code.control_code == Codes.WRONG_SCHEME:
+            ws_code_in_normal_scheme = True
+
+    # Check for a code in the WS code scheme
+    code_in_ws_scheme = False
+    ws_code = None
+    for label in coda_message.get_latest_labels():
+        if label.scheme_id == ws_code_scheme.scheme_id:
+            code_in_ws_scheme = True
+            ws_code = ws_code_scheme.get_code_with_code_id(label.code_id)
```
DanielMwendwa, line 124: Can this code block come before the `continue` statement in line 113?

src/engagement_db_coda_sync/lib.py
```diff
@@ -89,59 +156,56 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
...
-                message=engagement_db_message,
-                origin=HistoryEntryOrigin(origin_name="Coda -> Database Sync (WS Correction)", details=origin_details),
-                transaction=transaction
-            )
-
-            return
+    # WS-correct if there is a valid ws_code
+    if ws_code is not None:
+        try:
+            correct_dataset = coda_config.get_dataset_config_by_ws_code_string_value(ws_code.string_value).engagement_db_dataset
```
DanielMwendwa, line 172: My editor suggests we format this, its quite long.

DanielMwendwa (approved):

src/engagement_db_coda_sync/lib.py
as2388: It could, but I kept it separate because we're searching for two different things in each loop, and I was finding the logic was getting a bit tough to read if we did it all in one. What do you think?

DanielMwendwa: yeah, I get you, I think then it will be better if we update the condition on L112 to check for inequality & remove the continue statement.

as2388: agreed, will make that change instead, thanks

PR #51, opened 2021-07-20
Fixes fold_messages_by_uid not handling nested traced data correctly

as2388:
Fixes the problem by using append_traced_data rather than .serialize().

Also restructures to remove all the nested if/elses so we only have to understand a single code path.

### DIFF
--- a/src/engagement_db_to_analysis/engagement_db_to_analysis.py
+++ b/src/engagement_db_to_analysis/engagement_db_to_analysis.py
@@ -142,13 +142,13 @@ def _fold_messages_by_uid(user, messages_traced_data):
 
         # Get the existing list of messages for this dataset, if it exists, otherwise initialise with []
         participant_td = participants_traced_data_map[participant_uuid]
-        dataset_messages = participant_td.get(message_dataset, [])
+        participant_dataset_messages = participant_td.get(message_dataset, [])
 
         # Append this message to the list of messages for this dataset, and write-back to TracedData.
-        dataset_messages = dataset_messages.copy()
-        dataset_messages.append(dict(message))
+        participant_dataset_messages = participant_dataset_messages.copy()
+        participant_dataset_messages.append(dict(message))
         participant_td.append_data(
-            {message_dataset: dataset_messages},
+            {message_dataset: participant_dataset_messages},
             Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string())
         )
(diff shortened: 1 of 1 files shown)


src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -135,28 +135,32 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
 
-        if message["participant_uuid"] not in participants_traced_data_map.keys():
+        # Create an empty TracedData for this participant if this participant hasn't been seen yet.
+        if participant_uuid not in participants_traced_data_map.keys():
+            participants_traced_data_map[participant_uuid] = \
+                TracedData({}, Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string()))
+
+        # Get the existing list of messages for this dataset, if it exists, otherwise initialise with []
+        participant_td = participants_traced_data_map[participant_uuid]
+        dataset_messages = participant_td.get(message_dataset, [])
```
DanielMwendwa, line 145: I suggest renaming `dataset_messages` to sth like `dataset_messages_for_participant`

DanielMwendwa (approved):

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -135,28 +135,32 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+
+        # Append this message to the list of messages for this dataset, and write-back to TracedData.
+        dataset_messages = dataset_messages.copy()
+        dataset_messages.append(dict(message))
+        participant_td.append_data(
+            {message_dataset: dataset_messages},
+            Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string())
+        )
+        # TODO: Consider not converting messages to TracedData before this function, as that would remove the need
```
IsaackMwenda, line 154: unfortunately, we need to if we want to keep history for message filters.

IsaackMwenda (approved):

src/engagement_db_to_analysis/engagement_db_to_analysis.py
as2388: Thanks, renamed to `participant_dataset_messages`

src/engagement_db_to_analysis/engagement_db_to_analysis.py
as2388: converted to an explanation

PR #53, opened 2021-07-26
filter test uids at message td level.

IsaackMwenda:


src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -65,45 +65,41 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
     return filtered
 
 
-def filter_test_participants(user, participants_traced_data_map, test_participant_uuids):
+def filter_test_participants_messages(user, messages_traced_data, test_participant_uuids):
```
as2388, line 68: I think this can just be filter_test_messages(

src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -65,45 +65,41 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
...
+    filtered = []
 
-    for uid, participant_trace_data in participants_traced_data_map.items():
-        if uid in test_participant_uuids:
+    for td in messages_traced_data:
+        if td["participant_uuid"] in test_participant_uuids:
             continue
 
-        participant_trace_data.append_data(participant_trace_data, Metadata(user, Metadata.get_call_location(),
+        td.append_data(td, Metadata(user, Metadata.get_call_location(),
```
as2388, line 86: Can you add a comment here explaining why you're appending the td to itself, as it's not obvious on first reading

src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -65,45 +65,41 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
...
                                                     TimeUtils.utc_now_as_iso_string()))
-        filtered[uid] = participant_trace_data
+        filtered.append(td)
 
-    log.info(f"Returning {len(filtered)}/{len(participants_traced_data_map)} participants data...")
+    log.info(f"Returning {len(filtered)}/{len(filtered)} participants data...")
```
as2388, line 90:
```suggestion
    log.info(f"Returning {len(filtered)}/{len(messages_traced_data)} messages...")
```

as2388: I know this is filtering by message, but I wonder if it would be helpful to also print the number of removed participant uuids here too?

as2388 (approved):

src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -65,45 +65,41 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
     return filtered
 
 
-def filter_test_participants(user, participants_traced_data_map, test_participant_uuids):
+def filter_test_participants_messages(user, messages_traced_data, test_participant_uuids):
     """
     Filters a dict of participants who are not in pipeline_config.test_contacts e.g AVF/Aggregator staff
```
DanielMwendwa, line 70: `pipeline_config.test_contacts` -> `pipeline_config.test_participant_uuids`

src/engagement_db_to_analysis/traced_data_filters.py
```diff
@@ -65,45 +65,41 @@ def rqa_time_range_filter(user, messages_traced_data, pipeline_config):
...
     """
     log.debug("Filtering out test participants data...")
-    filtered = {}
+    filtered = []
 
-    for uid, participant_trace_data in participants_traced_data_map.items():
-        if uid in test_participant_uuids:
+    for td in messages_traced_data:
+        if td["participant_uuid"] in test_participant_uuids:
             continue
```
DanielMwendwa, line 84: You can write `if td["participant_uuid"] not in test_participant_uuids:` and avoid the continue statement.

src/engagement_db_to_analysis/traced_data_filters.py
DanielMwendwa: I suggest updating to sth like "Filters out test messages"

DanielMwendwa (approved):

src/engagement_db_to_analysis/traced_data_filters.py
IsaackMwenda: Yup, updated thanks.

src/engagement_db_to_analysis/traced_data_filters.py
IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/traced_data_filters.py
IsaackMwenda: Yup that would be useful. Added.

src/engagement_db_to_analysis/traced_data_filters.py
IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/traced_data_filters.py
IsaackMwenda: Here `continue` give us a convenient and clean strategy to log filtered out uuids and to skip to the next iteration right now. 
I don't see the benefit of _avoiding it_ .

DanielMwendwa: I think then it's better you add a log statement before the continue statement

PR #56, opened 2021-07-26
Handle runs from deleted contacts

as2388:
Hit this problem when testing https://github.com/AfricasVoices/Test-Pipeline-Engagement-DB/pull/55. In that case, I was able to verify that these were from deleted contacts by searching the raw messages (as these contain the original contact uuid with an anonymised urn that begins "deleted:")

### DIFF
--- a/src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
+++ b/src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
@@ -200,6 +200,13 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
                 continue
 
             # De-identify the contact's full urn.
+            if run.contact.uuid not in contacts_lut:
+                log.warning(f"Found a run from a contact that isn't present in the contacts export; skipping. "
+                            f"This is most likely because the contact was deleted, but could suggest a more serious "
+                            f"problem.")
+                if cache is not None:
+                    cache.set_latest_run_timestamp(flow_id, flow_config.flow_result_field, run.modified_on)
+                continue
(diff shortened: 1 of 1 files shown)


src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -200,6 +200,13 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
                 continue
 
             # De-identify the contact's full urn.
+            if run.contact.uuid not in contacts_lut:
+                log.warning(f"Found a run from a contact that isn't present in the contacts export; skipping. "
```
IsaackMwenda, line 204: assert if UUID is deleted so that we can monitor any other case.

IsaackMwenda (approved):

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
as2388: Unfortunately I don't think we can do this easily, as runs only contain the rapid pro contact uuid, which doesn't get marked as deleted. The only way to auto-detect is to download all the messages too, search those for uuids that match those not in contacts, then check that the urn starts with deleted. Do you think we should do that?

DanielMwendwa (approved):

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
IsaackMwenda: I see - let monitor the cases and reconsider doing this if we get other cases.

PR #58, opened 2021-07-27
Extract the coda -> db dataset sync logic into its own function

as2388:
This means the coda -> db sync now has the same structure as the db -> coda sync


DanielMwendwa (approved):
Great!

IsaackMwenda (approved):

PR #60, opened 2021-07-27
Make cache path optional

DanielMwendwa:


src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -9,30 +9,33 @@
 log = Logger(__name__)
 
 
-def _get_project_messages_from_engagement_db(analysis_configurations, engagement_db, cache_path):
+def _get_project_messages_from_engagement_db(analysis_configurations, engagement_db, cache_path=None):
     """
-
-    Downloads project messages from engagement database. It performs a full download if there is no previous export and
-    incrementally otherwise.
+    Gets project messages from engagement database.
```
as2388, line 14: I actually prefer the old description (with previous export replaced with cache path)

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -72,12 +75,14 @@ def _get_project_messages_from_engagement_db(analysis_configurations, engagement
...
-                cache.set_latest_message_timestamp(engagement_db_dataset, latest_message_timestamp)
+            if cache is not None:
+                # Export latest message timestamp to cache
+                if latest_message_timestamp is not None:
+                    cache.set_latest_message_timestamp(engagement_db_dataset, latest_message_timestamp)
 
-            # Export project engagement_dataset files
-            cache.set_messages(engagement_db_dataset, messages)
+                # Export project engagement_dataset files
+                if len(messages) > 0:
```
as2388, line 84: Why do we need this check?

as2388 (approved):

src/engagement_db_to_analysis/engagement_db_to_analysis.py
DanielMwendwa: I've updated the description, thanks.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
DanielMwendwa: To check if there are messages to be set in the analysis cache; Does not allow the opening of the cache if there is no data.

IsaackMwenda (approved):

PR #64, opened 2021-07-29
Impute age categories

IsaackMwenda:
TODO: plug-in sync stats in a future pr


src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
...
+    # Get the configurations for age and age_category analysis datasets
+    updated_messages_traced_data = []
+    age_cc = None
+    age_category_cc = None
+    age_engagement_db_datasets = None
+    for analysis_dataset_config in analysis_dataset_configs:
+        for coding_config in analysis_dataset_config.coding_configs:
+            if coding_config.analysis_dataset == "age":
+                age_cc = coding_config
+                age_engagement_db_datasets = analysis_dataset_config.engagement_db_datasets
```
DanielMwendwa, line 33: I suggest renaming `age_engagement_db_datasets` to `analysis_age_dataset_config` and then update L33 to `analysis_age_dataset_config = analysis_dataset_config` so that the code block does what you've commented in L24. We can then get age engagement db datasets by `analysis_age_dataset_config.engagement_db_datasets`

src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
...
+            elif coding_config.analysis_dataset == "age_category":
+                age_category_cc = coding_config
+
+    #Check and impute age_category in age messages only
+    for message in messages_traced_data:
```
DanielMwendwa, line 38: move `updated_messages_traced_data = []` on top of this for loop.

src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
...
+        if message["dataset"] not in age_engagement_db_datasets:
+            updated_messages_traced_data.append(message)
+            continue
```
DanielMwendwa, line 41: You can check for `message["dataset"] in age_engagement_db_datasets` and then update messages traced data at the end of the for loop.

src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+        # Append this age_category_label to the list of labels for this message, and write-back to TracedData.
+        message_labels = message["labels"].copy()
+        message_labels.append(age_category_label.to_dict())
+        message.append_data(
+            {"labels": message_labels},
+            Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string())
+        )
+
+        updated_messages_traced_data.append(message)
```
DanielMwendwa, line 73: Should we here also append trace data to message with message history?

DanielMwendwa (approved):

configurations/test_pipeline_configuration.py
```diff
@@ -142,7 +142,14 @@
                     ),
                     CodingConfiguration(
                         code_scheme=load_code_scheme("age_category"),
-                        analysis_dataset="age_category"
+                        analysis_dataset="age_category",
+                        age_categories = {
```
as2388, line 146:
```suggestion
                        age_categories={
```

configurations/test_pipeline_configuration.py
```diff
@@ -142,7 +142,14 @@
...
+                            (10, 14): "10 to 14",
+                            (15, 17): "15 to 17",
+                            (18, 35): "18 to 35",
+                            (36, 54): "36 to 54",
+                            (55, 99): "55 to 99"
+                        }
```
as2388, line 152: I think this is ok and happy to keep like this so we can keep making progress.

The slight hesitations I have are that
i. this means we need to specify the categories every time, rather than relying on an avf default.
ii. the functional programmer in me instinctively leans to something like code_imputations=[ImputeAgeCategories(categories=...), ...] as it makes adding additional imputations easier, but that might well be overkill here

src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+    :param user: Identifier of user running the pipeline.
+    :type user: str
+    :param messages_traced_data: Messages TracedData objects to impute age_category.
+    :type messages_traced_data: list of TracedData
+    :param analysis_dataset_config: Analysis dataset configuration in pipeline configuration module.
+    :type analysis_dataset_config: pipeline_config.analysis_configs.dataset_configurations
+    """
+
+    log.info("Imputing age categories for age dataset messages...")
```
as2388, line 23: I think we should say what the dataset actually is, as it might not exactly be called 'age'

src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
+from core_data_modules.util import TimeUtils
```
as2388, line 1: Can we call this file code_imputation_functions instead? I think all of analysis can arguably be described as a data wrangling operation

src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+    log.info("Imputing age categories for age dataset messages...")
+    # Get the configurations for age and age_category analysis datasets
+    updated_messages_traced_data = []
+    age_cc = None
+    age_category_cc = None
+    age_engagement_db_datasets = None
+    for analysis_dataset_config in analysis_dataset_configs:
+        for coding_config in analysis_dataset_config.coding_configs:
+            if coding_config.analysis_dataset == "age":
```
as2388, line 31: I'm not sure about this because it means we always need to call these datasets age and age_category, which might not be true and it's not clear when writing a configuration file that you need to do this.

I think I see the problem though, which is that the age_categorisation needs to know where to look to find the age dataset - maybe change the configuration to something like
age_category=AgeCategoryConfiguration(age_dataset="age", [categories=...])

src/engagement_db_to_analysis/data_wrangling_functions.py
```diff
@@ -0,0 +1,81 @@
...
+            age_category_code = age_category_cc.code_scheme.get_code_with_control_code(
+                age_code.control_code)
+
+        age_category_label = CleaningUtils.make_label_from_cleaner_code(
+            age_category_cc.code_scheme, age_category_code, Metadata.get_call_location()
+        )
+
+        # Append this age_category_label to the list of labels for this message, and write-back to TracedData.
+        message_labels = message["labels"].copy()
+        message_labels.append(age_category_label.to_dict())
```
as2388, line 67: Labels are sorted newest first, so I think you want to append at the front of this list

as2388 (changes requested):

configurations/test_pipeline_configuration.py
IsaackMwenda: on i)  We've seen this vary between projects in Cesc depending on the target audience. Moving this here standardizes _impute_age_category() so we update the categories & scheme between projects.
ii) My view is storing imputation configs in their resp coding_plan makes it easier to trace computations that happen in each analysis dataset.

src/engagement_db_to_analysis/data_wrangling_functions.py
IsaackMwenda: So we only need the coding_config and age_engagement_db_datasets for the rest of the script and not really analysis_age_dataset_config. So have updated the comment in L24 to be more clear :).

src/engagement_db_to_analysis/data_wrangling_functions.py
IsaackMwenda: Updated thanks.

src/engagement_db_to_analysis/data_wrangling_functions.py
IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/data_wrangling_functions.py
IsaackMwenda: Not really because;
1. updated_messages_traced_data is just a normal list that stores the traced data objects.
2. We've attached history for every message we have mutated in L67.

src/engagement_db_to_analysis/data_wrangling_functions.py
IsaackMwenda: updated thanks

src/engagement_db_to_analysis/data_wrangling_functions.py
IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/data_wrangling_functions.py
IsaackMwenda: It's not just that we also need to loop through age messages only, and to determine this we have to check for an age config property in each message. The alternative here is to define constants e.g AGE_CATEGORY = "age_category" in configuration.py that we can then use in the scripts in place of strings.

configurations/test_pipeline_configuration.py
```diff
@@ -142,9 +142,19 @@
...
+                        analysis_dataset="age_category",
+                        age_category_config=AgeCategoryConfiguration(age_analysis_dataset="age", #age_cc.analysis_dataset
+                                                                     categories={
+                                                                         (10, 14): "10 to 14",
+                                                                         (15, 17): "15 to 17",
+                                                                         (18, 35): "18 to 35",
+                                                                         (36, 54): "36 to 54",
+                                                                         (55, 99): "55 to 99"
+                                                                     }
+                                                                     )
```
as2388, line 154:
```suggestion
                        age_category_config=AgeCategoryConfiguration(
                            age_analysis_dataset="age",
                            categories={
                                (10, 14): "10 to 14",
                                (15, 17): "15 to 17",
                                (18, 35): "18 to 35",
                                (36, 54): "36 to 54",
                                (55, 99): "55 to 99"
                             }
                      )
```

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+    :param analysis_dataset_configs: Analysis dataset configuration in pipeline configuration module.
+    :type analysis_dataset_configs: pipeline_config.analysis_configs.dataset_configurations
+    """
+
+    # Get the coding configurations for age and age_category analysis datasets
+    age_category_cc = None
+    for analysis_dataset_config in analysis_dataset_configs:
+        for coding_config in analysis_dataset_config.coding_configs:
+            if coding_config.age_category_config is not None:
+                age_category_cc = coding_config
```
as2388, line 28: This will fail if we don't provide an age category dataset, and have unexpected results if we provide it twice. I think you need to return early (with a log) if there is no dataset where this was requested, and fail if you see it twice

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+    age_cc = None
```
as2388, line 30: Maybe this should be age_coding_config.

age_cc was fine when it was immediately preceded by for age_cc in coding_configurations:, but it's not obvious what the cc stands for without that for loop

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+    age_engagement_db_datasets = None
+    for analysis_dataset_config in analysis_dataset_configs:
+        for coding_config in analysis_dataset_config.coding_configs:
+            if coding_config.analysis_dataset == age_category_cc.age_category_config.age_analysis_dataset:
```
as2388, line 34: again, what happens if there were multiple matches. I think we need to check here (and/or pre-validate when the config gets loaded)

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+                age_cc = coding_config
+                age_engagement_db_datasets = analysis_dataset_config.engagement_db_datasets
+
+    # Check and impute age_category in age messages only
+    log.info(f"Imputing {age_category_cc.analysis_dataset} for {age_cc.analysis_dataset} messages...")
```
as2388, line 39:
```suggestion
    log.info(f"Imputing {age_category_cc.analysis_dataset} from {age_cc.analysis_dataset} messages...")
```

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+def _impute_age_category(user, messages_traced_data, analysis_dataset_configs):
+    """
+    Imputes age category for age dataset messages.
+
+    :param user: Identifier of user running the pipeline.
+    :type user: str
+    :param messages_traced_data: Messages TracedData objects to impute age_category.
+    :type messages_traced_data: list of TracedData
+    :param analysis_dataset_configs: Analysis dataset configuration in pipeline configuration module.
+    :type analysis_dataset_configs: pipeline_config.analysis_configs.dataset_configurations
```
as2388, line 20: Please doc the return (as with TracedData, we have to be super clear about when we're appending in place and when we're returning a modified copy)

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+            if coding_config.analysis_dataset == age_category_cc.age_category_config.age_analysis_dataset:
+                age_cc = coding_config
+                age_engagement_db_datasets = analysis_dataset_config.engagement_db_datasets
+
+    # Check and impute age_category in age messages only
+    log.info(f"Imputing {age_category_cc.analysis_dataset} for {age_cc.analysis_dataset} messages...")
+    updated_messages_traced_data = []
+    for message in messages_traced_data:
+        if message["dataset"] in age_engagement_db_datasets:
+            age_label = message["labels"][0]
```
as2388, line 43: What if we. have multiple labels assigned to this message? I think you need to get the latest label(s) with this code scheme - there is a function in column_view_conversion that handles this

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+            message_labels = message["labels"].copy()
+            message_labels.insert(0, age_category_label.to_dict())
+            message.append_data(
+                {"labels": message_labels},
+                Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string())
+            )
+
+        updated_messages_traced_data.append(message)
+
+        return updated_messages_traced_data
```
as2388, line 75: This means you're updating the message in place but returning a new list which is confusing. I think you either need to copy everything, which will be noticeably slow, or append in place, which I think is the better option here.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -159,6 +160,7 @@ def generate_analysis_files(user, pipeline_config, engagement_db, engagement_db_
 
     messages_traced_data = filter_messages(user, messages_traced_data, pipeline_config)
 
-    participants_traced_data_map = _fold_messages_by_uid(user, messages_traced_data)
+    messages_traced_data = run_code_computation_functions(user, messages_traced_data,
+                                                        pipeline_config.analysis_configs.dataset_configurations)
```
as2388, line 164:
```suggestion
    messages_traced_data = run_code_computation_functions(
        user, messages_traced_data, pipeline_config.analysis_configs.dataset_configurations
    )
```
In general I think this is the better way to indent, because otherwise you end up with lots of very short lines at the far right of the screen

src/pipeline_configuration_spec.py
```diff
@@ -45,11 +45,11 @@ class PipelineConfiguration:
...
     uuid_table: UUIDTableClientConfiguration
-    analysis_configs: AnalysisConfiguration
     project_start_date: datetime = None
     project_end_date: datetime = None
     test_participant_uuids: [] = None
     description: str = None
     rapid_pro_sources: [RapidProSource] = None
     coda_sync: CodaConfiguration = None
     rapid_pro_target: RapidProTarget = None
+    analysis_configs: AnalysisConfiguration = None
```
as2388, line 55: (drive-by comment, perhaps for another PR - This only holds one AnalysisConfiguration, so should be analysis_config not analysis_configs)

src/engagement_db_to_analysis/configuration.py
```diff
@@ -8,10 +9,16 @@ class DatasetTypes:
     DEMOGRAPHIC = "demographic"
     RESEARCH_QUESTION_ANSWER = "research_question_answer"
 
+@dataclass
+class AgeCategoryConfiguration:
+    age_analysis_dataset: str
+    categories: dict
+
```
as2388, line 16:
```suggestion


```
my editor is very noisy about PEP-8 violations without double-blank lines in modules

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -0,0 +1,81 @@
...
+
+def run_code_computation_functions(user, messages_traced_data, analysis_dataset_config):
```
as2388, line 77:
```suggestion
def impute_codes_by_message(user, messages_traced_data, analysis_dataset_config):
```
or similar - basically, we'll need to distinguish imputations that run message by message and imputations that run by participant

as2388 (approved):

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: Added an assert if age_category_cc is more than one and log for coding configs that don't have it.

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: updated to age_coding_config

src/engagement_db_to_analysis/engagement_db_to_analysis.py
IsaackMwenda: updated.

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: Updated thanks.

src/engagement_db_to_analysis/configuration.py
IsaackMwenda: Updated thanks.

src/pipeline_configuration_spec.py
IsaackMwenda: Thanks will update in a future PR.

PR #70, opened 2021-08-02
Convert analysis to column-view format

as2388:
As discussed offline, analysis needs to be (roughly) 3 phase:
1. Process message-by-message (incrementally get messages, filter irrelevant messages, impute codes etc.)
2. Convert the data format from a list of messages to two aggregated lists, one grouped by participant and the other grouped by rqa message.
3. Process the new formats (further impute codes, export analysis files, run automated analysis etc.)

This PR does the conversion (step 2). It also exports production files (sorry @DanielMwendwa for reimplementing this, but they trivially dropped out of the new format in 3 loc and were super useful here as a debug tool).

I think the format is right and will be easy to use with the existing automated analysis modules we have. It even uses the existing AnalysisConfiguration from core. My biggest issue here has been the terminology, for example we need to distinguish a message, message traced data, list of message traced data, and aggregated column-view rqa message traced data :/ If there's anything that's still confusing here we can think more about how to tighten this up.

### DIFF
--- a/src/engagement_db_to_analysis/engagement_db_to_analysis.py
+++ b/src/engagement_db_to_analysis/engagement_db_to_analysis.py
@@ -164,11 +164,12 @@ def _analysis_dataset_config_for_message(analysis_dataset_configs, message):
                      f"db dataset {message.dataset}")
 
 
-def _get_normalised_labels_for_code_scheme(message, code_scheme):
+def _get_latest_labels_with_code_scheme(message, code_scheme):
     """
-    Gets the labels assigned to this message under the given `code_scheme` (or a duplicate of this code scheme),
-    with label scheme_id's normalised to the primary code scheme id in cases where the code scheme was duplicated
-    e.g. scheme_id 'scheme-abc123-1' will be re-written to 'scheme-abc123'.
+    Gets the labels assigned to this message under the given `code_scheme` (or a duplicate of this code scheme).
+
+    Labels assigned under duplicate code schemes are normalised to have the primary code scheme id e.g. scheme_id
+    'scheme-abc123-1' will be re-written to 'scheme-abc123'.
 
     :param message: Message to get the labels from.
     :type message: engagement_database.data_models.Message
@@ -177,12 +178,12 @@ def _get_normalised_labels_for_code_scheme(message, code_scheme):
     :return: List of the relevant, normalised labels for the given code scheme.
     :rtype: list of core_data_modules.data_models.Label
     """
-    relevant_labels = []
+    latest_labels_with_code_scheme = []
     for label in message.get_latest_labels():
         if label.scheme_id.startswith(code_scheme.scheme_id):
             label.scheme_id = code_scheme.scheme_id
-            relevant_labels.append(label)
-    return relevant_labels
+            latest_labels_with_code_scheme.append(label)
+    return latest_labels_with_code_scheme
 
 
 def _add_message_to_column_td(user, message_td, column_td, analysis_config):
@@ -228,18 +229,18 @@ def _add_message_to_column_td(user, message_td, column_td, analysis_config):
     # If the TracedData already contains data here, combine the labels with the existing labels using
     # FoldStrategies.list_of_labels.
     for column_config in column_configs:
-        relevant_message_labels = _get_normalised_labels_for_code_scheme(message, column_config.code_scheme)
+        latest_labels_with_code_scheme = _get_latest_labels_with_code_scheme(message, column_config.code_scheme)
 
-        if len(relevant_message_labels) == 0:
+        if len(latest_labels_with_code_scheme) == 0:
             continue
 
-        relevant_message_labels = [label.to_dict() for label in relevant_message_labels]  # serialize for TracedData
+        latest_labels_with_code_scheme = [label.to_dict() for label in latest_labels_with_code_scheme]
         existing_labels = column_td.get(column_config.coded_field)
         if existing_labels is None:
-            new_data[column_config.coded_field] = relevant_message_labels
+            new_data[column_config.coded_field] = latest_labels_with_code_scheme
         else:
             new_data[column_config.coded_field] = FoldStrategies.list_of_labels(
-                column_config.code_scheme, existing_labels, relevant_message_labels
+                column_config.code_scheme, existing_labels, latest_labels_with_code_scheme
             )
 
     # Append the TracedData history for this message to the column-view.
--- a/src/engagement_db_to_analysis/engagement_db_to_analysis.py
+++ b/src/engagement_db_to_analysis/engagement_db_to_analysis.py
@@ -160,7 +160,7 @@ def _fold_messages_by_uid(user, messages_traced_data):
     return participants_traced_data_map
 
 
-def _analysis_dataset_config_to_column_config(analysis_dataset_config):
+def _analysis_dataset_config_to_column_configs(analysis_dataset_config):
     """
     Converts an analysis dataset configuration to the relevant "column-view" configurations.
 
@@ -253,7 +253,7 @@ def _add_message_to_column_td(user, message_td, column_td, analysis_config):
     analysis_dataset_config = _analysis_dataset_config_for_message(analysis_config.dataset_configurations, message)
 
     # Convert the analysis dataset config to its "column-view" configurations
-    column_configs = _analysis_dataset_config_to_column_config(analysis_dataset_config)
+    column_configs = _analysis_dataset_config_to_column_configs(analysis_dataset_config)
 
     new_data = dict()
 
@@ -404,6 +404,11 @@ def export_production_file(traced_data_iterable, analysis_config, export_path):
     with open(export_path, "w") as f:
         headers = ["participant_uuid"] + [c.raw_dataset for c in analysis_config.dataset_configurations]
         TracedDataCSVIO.export_traced_data_iterable_to_csv(traced_data_iterable, f, headers)
+
+
+def export_analysis_file(traced_data_iterable, analysis_config, export_path):
+    headers = ["participant_uuid"]
+    column_configs = analysis_config.dataset_configurations
 
 
 def export_traced_data(traced_data, export_path):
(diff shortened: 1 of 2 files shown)


src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+    participant, or by project.
+
+    :param analysis_dataset_config: Analysis dataset configuration to convert.
+    :type analysis_dataset_config: src.engagement_db_to_analysis.configuration.AnalysisDatasetConfiguration
+    :return: List of all the column configurations for this analysis dataset configuration.
+    :rtype: list of core_data_modules.analysis.analysis_utils.AnalysisConfiguration
+    """
+    column_configs = []
+    for coding_config in analysis_dataset_config.coding_configs:
+        config = AnalysisConfiguration(
```
IsaackMwenda, line 181:
```suggestion
        column_config = AnalysisConfiguration(
```

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+    e.g. scheme_id 'scheme-abc123-1' will be re-written to 'scheme-abc123'.
+
+    :param message: Message to get the labels from.
+    :type message: engagement_database.data_models.Message
+    :param code_scheme: Code scheme to get the latest labels for.
+    :type code_scheme: core_data_modules.data_models.CodeScheme
+    :return: List of the relevant, normalised labels for the given code scheme.
+    :rtype: list of core_data_modules.data_models.Label
+    """
+    relevant_labels = []
```
IsaackMwenda, line 223:
```suggestion
    message_labels = [] or normalised_labels
```
relevant is easily interpretable as normal codes only which is not the case.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+        new_data[analysis_dataset_config.raw_dataset] = message.text
+    else:
+        new_data[analysis_dataset_config.raw_dataset] = FoldStrategies.concatenate(existing_text, message.text)
+
+    # For each column config, get the latest, normalised labels under that column config's code scheme, and them
+    # to the column-view TracedData.
+    # If the TracedData already contains data here, combine the labels with the existing labels using
+    # FoldStrategies.list_of_labels.
+    for column_config in column_configs:
+        relevant_message_labels = _get_normalised_labels_for_code_scheme(message, column_config.code_scheme)
```
IsaackMwenda, line 274:
```suggestion
        normalised_message_labels = _get_normalised_labels_for_code_scheme(message, column_config.code_scheme)
```

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+    :rtype: list of core_data_modules.traced_data.TracedData
+    """
+    messages_by_column = dict()  # of participant_uuid -> list of rqa messages in column view
+
+    # Conduct the conversion in 2 passes.
+    # Pass 1: Convert each rqa message to a new TracedData object in column-view.
+    for msg_td in messages_traced_data:
+        # Skip this message if it's not an RQA
+        message = Message.from_dict(dict(msg_td))
+        analysis_dataset_config = _analysis_dataset_config_for_message(analysis_config.dataset_configurations, message)
```
IsaackMwenda, line 321:
```suggestion
        message_analysis_dataset_config = _analysis_dataset_config_for_message(analysis_config.dataset_configurations, message)
```

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+            participants_by_column[message.participant_uuid] = TracedData(
+                {"participant_uuid": message.participant_uuid},
+                Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string())
+            )
+
+        # Add this message to the relevant participant's column-view TracedData.
+        participant = participants_by_column[message.participant_uuid]
+        _add_message_to_column_td(user, msg_td, participant, analysis_config)
+
+    return participants_by_column.values()
```
IsaackMwenda, line 387: for consistency, lets return a normal list of participants td instead of values() view object list.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -164,6 +424,14 @@ def generate_analysis_files(user, pipeline_config, engagement_db, cache_path=Non
 
     messages_traced_data = filter_messages(user, messages_traced_data, pipeline_config)
 
-    participants_traced_data_map = _fold_messages_by_uid(user, messages_traced_data)
+    messages_by_column = _convert_to_messages_column_format(user, messages_traced_data, pipeline_config.analysis_configs)
```
IsaackMwenda, line 427: Does it make sense to move the other helper functions to a traced_data_conversion library? That way we can reuse some of them in other scripts e.g `_analysis_dataset_config_for_message`....

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -164,6 +424,14 @@ def generate_analysis_files(user, pipeline_config, engagement_db, cache_path=Non
...
+    participants_by_column = _convert_to_participants_column_format(user, messages_traced_data, pipeline_config.analysis_configs)
```
IsaackMwenda, line 428: Please remove _fold_messages_by_uid function in [L117] as this is no longer relevant.

IsaackMwenda (changes requested):
Thanks for the great progress here @as2388 . Requesting changes so that have a final look at the suggestion before will merge this.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+    :param messages_traced_data: Messages traced data to convert.
+    :type messages_traced_data: list of core_data_modules.traced_data.TracedData
+    :param analysis_config: Configuration for the conversion.
+    :type analysis_config: src.engagement_db_to_analysis.configuration.AnalysisConfiguration
+    :return: Messages organised by participant into column-view format  suitable for further analysis.
+    :rtype: list of core_data_modules.traced_data.TracedData
+    """
+    participants_by_column = dict()  # of participant_uuid -> participant traced data in column view
+    for msg_td in messages_traced_data:
+        message = Message.from_dict(dict(msg_td))
```
DanielMwendwa, line 374: consider getting `analysis_dataset_config = _analysis_dataset_config_for_message(analysis_config.dataset_configurations, message)` after L374 so that `_add_message_to_column_td` accepts `analysis_dataset_config` as an arg instead of `analysis_config` for consistency

src/engagement_db_to_analysis/engagement_db_to_analysis.py
as2388: agreed, updated

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+    """
+    message = Message.from_dict(dict(message_td))
+
+    # Get the analysis dataset configuration for this message
+    analysis_dataset_config = _analysis_dataset_config_for_message(analysis_config.dataset_configurations, message)
+
+    # Convert the analysis dataset config to its "column-view" configurations
+    column_configs = _analysis_dataset_config_to_column_configs(analysis_dataset_config)
+
+    new_data = dict()
```
DanielMwendwa, line 258: I think we should rename `new_data` describing it as a column

DanielMwendwa (approved):
Thanks, @as2388.  You may consider splitting  `_add_message_to_column_td` into two, we can have a function that returns the `new_data`  and let `_add_message_to_column_td` accept the `new_data` .

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -156,6 +160,262 @@ def _fold_messages_by_uid(user, messages_traced_data):
...
+     - Labels are handled by `core_data_modules.util.fold_traced_data.FoldStrategies.list_of_labels`.
+
+    :param user: Identifier of user running the pipeline.
+    :type user: str
+    :param message_td: TracedData representing the message to add
+    :type message_td: core_data_modules.traced_data.TracedData
+    :param column_td: An existing TracedData object in column-view format, to which the relevant data from this message
+                      will be appended.
+    :type column_td: core_data_modules.traced_data.TracedData
+    :param analysis_config: Analysis configuration.
```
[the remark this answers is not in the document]
as2388: hmm I think this should be analysis_config because it only accepts one AnalysisConfiguration.

src/engagement_db_to_analysis/engagement_db_to_analysis.py
as2388: thanks, forgot py 3 created views

src/engagement_db_to_analysis/engagement_db_to_analysis.py
as2388: agreed. I think better to do this in a follow-up PR so it's easier to review the move

src/engagement_db_to_analysis/engagement_db_to_analysis.py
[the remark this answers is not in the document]
IsaackMwenda: I think we accept all here then compute the one that the message belongs to in l211.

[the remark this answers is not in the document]
as2388: I think the confusion here and at https://github.com/AfricasVoices/Test-Pipeline-Engagement-DB/pull/70/files#r682720732 is because the function is accepting a single, top-level AnalysisConfiguration but only using the AnalysisConfiguration.DatasetConfigurations. I'll change this to take arg analysis_dataset_configs instead

IsaackMwenda (approved):

PR #74, opened 2021-08-09
Add Dockerfile

DanielMwendwa:


Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# Install Python tools (git + pipenv)
+RUN apt-get update && apt-get install -y git
+RUN pip install pipenv
+
+# Install memory_profiler if this script is run with PROFILE_MEMORY flag
+ARG INSTALL_MEMORY_PROFILER="false"
+RUN if [ "$INSTALL_MEMORY_PROFILER" = "true" ]; then \
+        apt-get update && apt-get install -y gcc && \
+        pip install memory_profiler; \
+    fi
```
as2388, line 12: Do we need still need this cc @IsaackMwenda. Unless we're really sure that we do, I'm hesitant to add it now

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN pipenv sync
+
+# Copy the rest of the project
+ADD code_schemes/*.json /app/code_schemes/
+ADD configurations/ /app/configurations/
+ADD src /app/src
+ADD sync_coda_to_engagement_db.py /app
+ADD sync_engagement_db_to_coda.py /app
+ADD sync_engagement_db_to_rapid_pro.py /app
+ADD engagement_db_to_analysis.py /app
```
as2388, line 35: Is it ok to simplify this to ADD *.py /app?

as2388, line 35: It would probably have been a good idea to dockerise one of the steps at the same time so you could test this before sending for review. I think this will fail when dockerised because we now need the .git repo to be available too, which either means copying this folder too or allowing us to pass git info via arguments

as2388 (approved):

Dockerfile
IsaackMwenda: We can clean up for now and add on demand.

IsaackMwenda (approved):

Dockerfile
DanielMwendwa: removed

Dockerfile
DanielMwendwa: I think it is better to specify the scripts needed to be run using docker.

DanielMwendwa: I'll be dockerising one of the steps in a new PR.

DanielMwendwa: I will add the entire repository.

PR #76, opened 2021-08-10
impute age category in-place

IsaackMwenda:


src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -37,20 +39,25 @@ def _impute_age_category(user, messages_traced_data, analysis_dataset_configs):
...
+                assert age_coding_config is None, f"Found more than one age_coding_config in analysis_dataset_config"
                 age_coding_config = coding_config
                 age_engagement_db_datasets = analysis_dataset_config.engagement_db_datasets
 
     # Check and impute age_category in age messages only
-    log.info(f"Imputing {age_category_cc.analysis_dataset} from {age_coding_config.analysis_dataset} messages...")
-    for message_index, message in enumerate(messages_traced_data):
+    log.info(f"Imputing {age_category_cc.analysis_dataset} labels from {age_coding_config.analysis_dataset} messages...")
+    imputed_labels = 0
+    age_messages = 1
```
as2388, line 50: Why are these numbers different?

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -37,20 +39,25 @@ def _impute_age_category(user, messages_traced_data, analysis_dataset_configs):
...
+    for message in messages_traced_data:
         if message["dataset"] in age_engagement_db_datasets:
-            age_label = message["labels"][0]
-            age_code = age_coding_config.code_scheme.get_code_with_code_id(age_label["CodeID"])
+
+            age_labels = get_latest_labels_with_code_scheme(Message.from_dict(dict(message)), age_coding_config.code_scheme)
```
as2388, line 54: Maybe you could deserialize earlier so you can say things like if message.dataset in age_engagement_db_datasets

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -74,12 +81,25 @@ def _impute_age_category(user, messages_traced_data, analysis_dataset_configs):
...
 
-            messages_traced_data[message_index] = message
+            age_messages +=1
+            imputed_labels +=1
+
+    log.info(f"Imputed {imputed_labels} age category labels for {age_messages} age messages")
 
-        return messages_traced_data
 
 def impute_codes_by_message(user, messages_traced_data, analysis_dataset_config):
```
as2388, line 90: I think this should be analysis_dataset_configs plural? That's what the docstring says...

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -37,20 +39,25 @@ def _impute_age_category(user, messages_traced_data, analysis_dataset_configs):
...
+
+                assert age_coding_config is None, f"Found more than one age_coding_config in analysis_dataset_config"
                 age_coding_config = coding_config
                 age_engagement_db_datasets = analysis_dataset_config.engagement_db_datasets
 
     # Check and impute age_category in age messages only
-    log.info(f"Imputing {age_category_cc.analysis_dataset} from {age_coding_config.analysis_dataset} messages...")
-    for message_index, message in enumerate(messages_traced_data):
+    log.info(f"Imputing {age_category_cc.analysis_dataset} labels from {age_coding_config.analysis_dataset} messages...")
+    imputed_labels = 0
```
as2388, line 49: doesn't need to be addressed here, but this suggests that impute_age_labels, label_imputation_functions etc. is the correct terminology

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: This was an error. Thanks for catching it fixed.

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: I prefer deserializing on-demand here as the format is not used anywhere else in this function.

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: updated thanks.

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: updated to `log.info(f"Imputing {age_category_cc.analysis_dataset} labels **for** {age_coding_config.analysis_dataset} messages...")`

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: message["dataset"] is used on L52, and you get slightly better editor support with message.dataset

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -74,12 +81,25 @@ def _impute_age_category(user, messages_traced_data, analysis_dataset_configs):
                 Metadata(user, Metadata.get_call_location(), TimeUtils.utc_now_as_iso_string())
             )
 
-            messages_traced_data[message_index] = message
+            age_messages +=1
+            imputed_labels +=1
```
as2388, line 85: I think this is confusing because if we're tracking two variables then surely they should be tracking different things/be able to reach different numbers?

I think the numbers we care about are: (i) how many messages did you look at, and (ii) how many got imputed

as2388 (approved):

src/engagement_db_to_analysis/code_imputation_functions.py
IsaackMwenda: This was more of a validation log that we have imputed the same number of labels for the available age messages. Have moved age_messages += 1 to increment before imputing
        `if message["dataset"] in age_engagement_db_datasets:
            age_messages += 1`
but happy to log the number of messages traced data if you think that's more useful.

PR #84, opened 2021-08-24
Dockerise rapid pro to engagement db 

DanielMwendwa:


docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,85 @@
...
+if [[ $# -ne 3 ]]; then
+    echo "Usage: ./docker-run-sync-rapid-pro-to-engagement-db.sh
+    [--incremental-cache-path <incremental-cache-path>] [--local-archive <local_archive>]
+    <user> <google-cloud-credentials-file-path> <configuration-module>"
+    exit
+fi
+
+# Assign the program arguments to bash variables.
+USER=$1
+INPUT_GOOGLE_CLOUD_CREDENTIALS=$2
```
as2388, line 36:
```suggestion
GOOGLE_CLOUD_CREDENTIALS_PATH=$2
```
(I know you copied this from before, but I think this is better)

docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,85 @@
...
+set -e
+
+PROJECT_NAME="$(<configurations/docker_image_project_name.txt)"
+IMAGE_NAME=$PROJECT_NAME-rapid-pro-to-engagement-db
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --incremental-cache-path)
+            INCREMENTAL_MODE=true
+            INCREMENTAL_CACHE_PATH="$2"
```
as2388, line 12: I think in dockerised mode this should be a volume instead (and then we automatically create the appropriate sub-directory in the volume in this script)

docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,85 @@
...
+fi
+
+container="$(docker container create -w /app "$IMAGE_NAME" /bin/bash -c "$CMD")"
+echo "Created container $container"
+container_short_id=${container:0:7}
+
+# Copy input data into the container
+echo "Copying $INPUT_GOOGLE_CLOUD_CREDENTIALS -> $container_short_id:/credentials/google-cloud-credentials.json"
+docker cp "$INPUT_GOOGLE_CLOUD_CREDENTIALS" "$container:/credentials/google-cloud-credentials.json"
+
```
as2388, line 75: If there's a local archive you'll need to copy this in too

docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,85 @@
...
+function finish {
+    # Tear down the container when done.
+    docker container rm "$container" >/dev/null
+}
+trap finish EXIT
```
as2388, line 80: Can you replace this with just
`docker container rm "$container" >/dev/null`
at the very end of the script?

The trap finish structure always destroys the container, even if it crashes, which seemed like a good idea but it's actually turned out to be better to only delete when they run successfully, so that we can still access the data in the container that crashed to help debug

docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,85 @@
...
+# Create a container from the image that was just built.
+if [[ "$OPTIONAL_ARGS" ]]; then
+    CMD="pipenv run python -u sync_rapid_pro_to_engagement_db.py \"$OPTIONAL_ARGS\" \
+    \"$USER\" /credentials/google-cloud-credentials.json \"$INPUT_CONFIGURATION_MODULE\"
+    "
+else
+    CMD="pipenv run python -u sync_rapid_pro_to_engagement_db.py \"$USER\" \
+    /credentials/google-cloud-credentials.json \"$INPUT_CONFIGURATION_MODULE\"
+    "
+fi
```
as2388, line 66: You might be able to simplify this to just the if true case if you initialise OPTIONAL_ARGS to "" earlier in the script

as2388 (changes requested):

docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,84 @@
...
+
+set -e
+
+PROJECT_NAME="$(<configurations/docker_image_project_name.txt)"
+IMAGE_NAME=$PROJECT_NAME-rapid-pro-to-engagement-db
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --incremental-cache-volume)
+            INCREMENTAL_ARG=" --incremental-cache-volume /cache"
```
as2388, line 11:
```suggestion
            INCREMENTAL_ARG=" --incremental-cache-path /cache"
```

docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
```
as2388, line 1: I think this can be called just docker-sync-rapid-pro-to-engagement-db i.e. I think -run- is unnecessary

as2388, line 1: can you chmod +x this file please so I can $ ./docker-run...

docker-run-sync-rapid-pro-to-engagement-db.sh
```diff
@@ -0,0 +1,84 @@
...
+fi
+
+echo "Created container $container"
+container_short_id=${container:0:7}
+
+# Copy input data into the container
+echo "Copying $GOOGLE_CLOUD_CREDENTIALS_PATH -> $container_short_id:/credentials/google-cloud-credentials.json"
+docker cp "$GOOGLE_CLOUD_CREDENTIALS_PATH" "$container:/credentials/google-cloud-credentials.json"
+
+for LOCAL_ARCHIVE_PATH in "${LOCAL_ARCHIVE_PATHS[@]}"; do
```
as2388, line 67: Sorry you had to go through this. Well done finding a solution.

as2388 (approved):

docker-run-sync-rapid-pro-to-engagement-db.sh
DanielMwendwa: Agreed & updated.

DanielMwendwa: updated.

docker-run-sync-rapid-pro-to-engagement-db.sh
DanielMwendwa: Thanks.

IsaackMwenda (approved):
Reviewed offline - looks good now thanks

PR #88, opened 2021-08-27
Impute Codes.NOT_INTERNALLY_CONSISTENT for demogs

as2388:
Now that we're processing all demog messages from a person per question (rather than trying to pretend there was only one, like we tried to do in the past), we need to be careful not to interpret someone who sent conflicting answers (e.g. both "20" and "22") twice. Instead, this PR replaces those conflicts to an NIC label, to show that there is a problem with these demographics.

Tested on WorldVision data and confirmed that it's imputing NIC in the places we'd expect. It impacted 5/1287 individuals. Please check the logic extra carefully though as it's new for the engagement db pipelines.

### DIFF
--- a/src/engagement_db_to_analysis/code_imputation_functions.py
+++ b/src/engagement_db_to_analysis/code_imputation_functions.py
@@ -1,20 +1,16 @@
 from core_data_modules.cleaners import Codes
-from core_data_modules.util import TimeUtils
-from core_data_modules.data_models.code_scheme import CodeTypes
 from core_data_modules.cleaners.cleaning_utils import CleaningUtils
-from core_data_modules.traced_data import Metadata
-from core_data_modules.logging import Logger
-from core_data_modules.cleaners import Codes
 from core_data_modules.cleaners.location_tools import KenyaLocations
-
-
+from core_data_modules.data_models.code_scheme import CodeTypes
+from core_data_modules.logging import Logger
+from core_data_modules.traced_data import Metadata
+from core_data_modules.util import TimeUtils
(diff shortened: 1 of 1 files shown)


src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -333,12 +329,111 @@ def _impute_true_missing(user, column_traced_data_iterable, analysis_dataset_con
...
+    :type analysis_dataset_configs: pipeline_config.analysis_configs.dataset_configurations
+    """
+    imputed_codes = 0
+    log.info(f"Imputing {Codes.NOT_INTERNALLY_CONSISTENT} codes...")
+
+    demog_column_configs = _get_demog_column_configs(analysis_dataset_configs)
+    for td in column_traced_data_iterable:
+        for column_config in demog_column_configs:
+            if _demog_has_conflicting_normal_labels(td, column_config):
+                # Replace the conflicting normal labels with a NOT_INTERNALLY_CONSISTENT label,
```
IsaackMwenda, line 410: Should we entirely replace or impute NIC, retain Meta/Control codes, and keep conflicting labels history?

IsaackMwenda (approved):

DanielMwendwa (approved):

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: discussed offline, will keep as-is as we're not working with coda style datasets here so don't need to mix history with data

PR #92, opened 2021-08-30
Add automated analysis maps

as2388:
This is a significant improvement in implementation simplicity over the old version, thanks to the new analysis configuration format.

I think we'll want to rename KenyaAnalysisLocations -> AnalysisLocations so we can easily slot in the Somalia maps to in future. We might want to add additional configuration too for controlling how some of the maps are drawn, but until we hit such needs I think we should proceed with this convention-over-configuration case.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-CoreDataModules = {editable = true,git = "https://www.github.com/AfricasVoices/CoreDataModules",ref = "master"}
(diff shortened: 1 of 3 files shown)


DanielMwendwa (approved):

IsaackMwenda (approved):
LGTM, we can rename `KenyaAnalysisLocations.COUNTY` -> `AnalysisLocations.KENYA_COUNTY`

as2388:
> LGTM, we can rename `KenyaAnalysisLocations.COUNTY` -> `AnalysisLocations.KENYA_COUNTY`

agreed, I'll open a new pr for this shortly

PR #111, opened 2021-09-24
Add pipeline monitoring and end to end run script

IsaackMwenda:


as2388 (approved):

docker-run-log-pipeline-event.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+set -e
+
+PROJECT_NAME="$(<configurations/docker_image_project_name.txt)"
+IMAGE_NAME=$PROJECT_NAME-log-pipeline-event
+
+# Check that the correct number of arguments were provided.
+if [[ $# -ne 4 ]]; then
+    echo "Usage: ./docker-run-log-pipeline-event.sh
```
DanielMwendwa, line 10: You can use `echo "Usage: $0"` like in other scripts

log_pipeline_event.py
```diff
@@ -0,0 +1,49 @@
+import argparse
+import json
+import importlib
+
+
```
DanielMwendwa, line 5:
```suggestion
```

log_pipeline_event.py
```diff
@@ -0,0 +1,49 @@
...
+from core_data_modules.util import TimeUtils
+from storage.google_cloud import google_cloud_utils
+from pipeline_logs.firestore_pipeline_logger import FirestorePipelineLogger
+
+from src.common.configuration import PipelineEvents
+
+log = Logger(__name__)
+
+def log_pipeline_event(pipeline_config, google_cloud_credentials_file_path,  run_id, event_key):
+
```
DanielMwendwa, line 16:
```suggestion
```

DanielMwendwa (approved):

run_pipeline.sh
```diff
@@ -0,0 +1,38 @@
...
+./docker-sync-rapid-pro-to-engagement-db.sh --incremental-cache-volume "cache" "$USER" "$GOOGLE_CLOUD_CREDENTIALS_PATH" \
+                                    "$CONFIGURATION_MODULE"
+
+./docker-sync-engagement-db-to-coda.sh --incremental-cache-volume "cache" "$USER" "$GOOGLE_CLOUD_CREDENTIALS_PATH" \
+                        "$CONFIGURATION_MODULE"
+
+./docker-sync-coda-to-engagement-db.sh --incremental-cache-volume "cache" "$USER" "$GOOGLE_CLOUD_CREDENTIALS_PATH" \
+                        "$CONFIGURATION_MODULE"
+
+./docker-sync-engagement-db-to-rapid-pro.sh "$USER" "$GOOGLE_CLOUD_CREDENTIALS_PATH" "$CONFIGURATION_MODULE"
```
as2388, line 33: I don't think this is ready for realtime sync. I think for the first project we want to run this once at the start then leave it alone

docker-run-log-pipeline-event.sh
IsaackMwenda: What's the advantage of echo "Usage: $0" i think the latter is more descriptive?

run_pipeline.sh
```diff
@@ -0,0 +1,38 @@
...
+
+./docker-run-engagement-db-to-analysis.sh --incremental-cache-volume "cache" "$USER" "$GOOGLE_CLOUD_CREDENTIALS_PATH" \
+                        "$CONFIGURATION_MODULE" "$OUTPUT_DIR"
```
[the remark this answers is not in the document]
IsaackMwenda: Have added a PIPELINE_NAME arg in run_pipeline.sh and renamed the various stages cache names to "$PIPELINE_NAME-stage-name". This resolves the above constraints neatly.

run_pipeline.sh
IsaackMwenda: I agree, removed.

docker-run-log-pipeline-event.sh
DanielMwendwa: $0 expands to the name of the shell script thus helps in case of the script is renamed and the usage statement not updated. Also for consistency with other shell scripts.

PR #112, opened 2021-09-28
Impute WS coding errors in analysis

as2388:
(Running this branch over the WV data with WS correction disabled now gives the same results in engagement_counts.csv as the previous infrastructure)

### DIFF
--- a/src/engagement_db_to_analysis/code_imputation_functions.py
+++ b/src/engagement_db_to_analysis/code_imputation_functions.py
@@ -1,6 +1,7 @@
 from core_data_modules.cleaners import Codes
 from core_data_modules.cleaners.cleaning_utils import CleaningUtils
 from core_data_modules.cleaners.location_tools import KenyaLocations
+from core_data_modules.data_models import Label, Origin
 from core_data_modules.data_models.code_scheme import CodeTypes
 from core_data_modules.logging import Logger
 from core_data_modules.traced_data import Metadata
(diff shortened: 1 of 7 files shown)


src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -81,6 +82,105 @@ def _impute_not_reviewed_labels(user, messages_traced_data, analysis_dataset_con
...
+                continue
+
+            if label.scheme_id == ws_correct_dataset_code_scheme.scheme_id:
+                code_in_ws_scheme = True
+
+        if ws_code_in_normal_scheme != code_in_ws_scheme:
+            imputed_labels += 1
+
+            # Clear all duplicate schemes
+            valid_code_scheme_ids = [code_scheme.scheme_id for code_scheme in normal_code_schemes]
```
IsaackMwenda, line 156:
```suggestion
            valid_code_scheme_ids = set(normal_codes_schemes)
```
since we are storing unique randomly ordered ids.

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -81,6 +82,105 @@ def _impute_not_reviewed_labels(user, messages_traced_data, analysis_dataset_con
...
+            valid_code_scheme_ids.append(ws_correct_dataset_code_scheme.scheme_id)
+            for label in message.get_latest_labels():
+                cleared_label = None
+                for scheme_id in valid_code_scheme_ids:
+                    if label.scheme_id.startswith(scheme_id):
```
IsaackMwenda, line 161: Sorry, I might have missed this - but why are we inserting a cleared_label here?

IsaackMwenda (approved):

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: The messages at this point contain labels in Coda's format. If we want to say that these labels are invalid and a coding error, we need to replace the currently assigned labels across all schemes with Coding Error, taking care to clear pre-existing codes in duplicated schemes, in the same way that a human labeller would clear codes in Coda. Clearing all the labels here ensures that only the new labels we write will be returned by latest_labels() later on

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -81,6 +82,105 @@ def _impute_not_reviewed_labels(user, messages_traced_data, analysis_dataset_con
...
+    """
+    Imputes Codes.CODING_ERROR labels for messages that have a coding error in the WS labels that have been applied.
+
+    We consider WS labels to have a coding error if either of these conditions holds:
+     - There is a WS label  applied in a normal code scheme, but there is no label in the WS - Correct Dataset code scheme.
+     - There is a label applied in the WS - Correct Dataset code scheme, but no WS code in any of the normal code schemes.
+
+    :param user: Identifier of user running the pipeline.
+    :type user: str
+    :param messages_traced_data: Messages TracedData objects to impute not reviewed labels.
```
DanielMwendwa, line 117:
```suggestion
    :param messages_traced_data: Messages TracedData objects to impute ws coding errors.
```

DanielMwendwa (approved):

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: thanks, fixed

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: added an explanation as a comment in the code

PR #114, opened 2021-09-29
Fix NR edge cases

as2388:
Review #112 first.

This makes the following changes:
 - The script now checks the WS - Correct Dataset code scheme too when looking for review/not reviewed labels
 - The script now checks if all labels have been reviewed - if there is a mix of reviewed and not reviewed labels for the same label, then converts to coding error
 - Clears all existing labels before assigning NR labels, including across duplicated code schemes
 - Writes NR/CE to all code schemes, not just to the first.
 - Improves logging


IsaackMwenda (approved):
Thanks, handles messages with both checked + unchecked labels well.

DanielMwendwa (approved):

PR #119, opened 2021-10-13
Add scripts and configurations to support data dir backup to g-cloud

IsaackMwenda:


upload_backup_files.py
```diff
@@ -0,0 +1,130 @@
+import argparse
```
as2388, line 1: Would it make sense to do this at the end of the analysis script now that (i) it's only uploading data produced by analysis, rather than uploading data produced by many stages, and (ii) we've much improved the reliability of our uploads and no longer need to manually upload any that failed?

as2388 (approved):
code lgtm, just some questions around where the best place to run it is

upload_backup_files.py
IsaackMwenda: This actually uploads data from the cache in the other stages so that we can reproduce the analysis outputs without needing to fetch data from either Textit, Coda or Engagement Db. So inclined to separate it from analysis upload.

PR #120, opened 2021-10-25
Support membership groups

IsaackMwenda:
This adds support for membership groups e.g listening/ beneficiary groups with an optional configuration for projects that don't have them.

### DIFF
--- a/configurations/test_pipeline_configuration.py
+++ b/configurations/test_pipeline_configuration.py
@@ -98,6 +98,12 @@ PIPELINE_CONFIGURATION = PipelineConfiguration(
             credentials_file_url="gs://avf-credentials/pipeline-runner-service-acct-avf-data-core-64cc71459fe7.json",
             drive_dir="pipeline_upload_test"
         ),
+        membership_group_configuration=MembershipGroupConfiguration(
+            membership_group_csv_urls={ "listening_group": [
+                "gs://avf-project-datasets/2021/TEST-PIPELINE-ENGAGEMENT-DB/test-pipeline-engagement-db-listening-group.csv"
+            ]
+            },
+        ),
         dataset_configurations=[
             AnalysisDatasetConfiguration(
                 engagement_db_datasets=["s01e01"],
@@ -169,5 +175,5 @@ PIPELINE_CONFIGURATION = PipelineConfiguration(
     archive_configuration = ArchiveConfiguration(
         archive_upload_bucket = "gs://pipeline-execution-backup-archive",
         bucket_dir_path =  "2021/TEST-PIPELINE_DB"
-        )
+    )
 )
--- a/engagement_db_to_analysis.py
+++ b/engagement_db_to_analysis.py
@@ -12,6 +12,7 @@ if __name__ == "__main__":
 
     parser.add_argument("--incremental-cache-path",
                         help="Path to a directory to use to cache results needed for incremental operation.")
+
     parser.add_argument("user", help="Identifier of the user launching this program")
     parser.add_argument("google_cloud_credentials_file_path", metavar="google-cloud-credentials-file-path",
                         help="Path to a Google Cloud service account credentials file to use to access the "
@@ -19,6 +20,8 @@ if __name__ == "__main__":
     parser.add_argument("configuration_module",
                         help="Configuration module to import e.g. 'configurations.test_config'. "
                              "This module must contain a PIPELINE_CONFIGURATION property")
+    parser.add_argument("membership_group_dir_path",
+                        help="Path to a directory to use to read & write membership group csvs to.")
     parser.add_argument("output_dir", metavar="output-dir",
                         help="Directory to output all analysis results to. This script will create and organise the "
                              "outputs into sensible sub-directories automatically")
@@ -29,6 +32,7 @@ if __name__ == "__main__":
     user = args.user
     google_cloud_credentials_file_path = args.google_cloud_credentials_file_path
     pipeline_config = importlib.import_module(args.configuration_module).PIPELINE_CONFIGURATION
+    membership_group_dir_path = args.membership_group_dir_path
     output_dir = args.output_dir
 
     pipeline = pipeline_config.pipeline_name
@@ -36,6 +40,5 @@ if __name__ == "__main__":
     uuid_table = pipeline_config.uuid_table.init_uuid_table_client(google_cloud_credentials_file_path)
     engagement_db = pipeline_config.engagement_database.init_engagement_db_client(google_cloud_credentials_file_path)
 
-    generate_analysis_files(
-        user, google_cloud_credentials_file_path, pipeline_config, engagement_db, output_dir, incremental_cache_path
-    )
+    generate_analysis_files(user, google_cloud_credentials_file_path, pipeline_config, engagement_db, membership_group_dir_path,
+    output_dir, incremental_cache_path)
--- a/src/engagement_db_to_analysis/analysis_files.py
+++ b/src/engagement_db_to_analysis/analysis_files.py
@@ -30,25 +30,33 @@ def export_production_file(traced_data_iterable, analysis_config, export_path):
         TracedDataCSVIO.export_traced_data_iterable_to_csv(traced_data_iterable, f, headers)
 
 
-def _get_analysis_file_headers(column_configs, export_timestamps=False):
+
+def _get_analysis_file_headers(pipeline_config, export_timestamps=False):
     """
     Gets the headers for an analysis file.
 
     The headers are:
      - "participant_uuid"
      - "consent_withdrawn"
+     - Membership_group e.g "listening_group" - Optional
      - Labels for each normal code scheme, in matrix format e.g. "age:25", "s01e01:healthcare".
      - Raw messages for each dataset.
-
-    :param column_configs: Column configurations to use to derive the headers.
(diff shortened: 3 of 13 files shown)


configurations/test_pipeline_configuration.py
```diff
@@ -98,6 +98,12 @@
             credentials_file_url="gs://avf-credentials/pipeline-runner-service-acct-avf-data-core-64cc71459fe7.json",
             drive_dir="pipeline_upload_test"
         ),
+        membership_group_configuration=MembershipGroupConfiguration(
+            membership_group_csv_urls={ "listening_group": [
+                "gs://avf-project-datasets/2021/TEST-PIPELINE-ENGAGEMENT-DB/test-pipeline-engagement-db-listening-group.csv"
+            ]
```
DanielMwendwa, line 104: tab indent.

engagement_db_to_analysis.py
```diff
@@ -12,13 +12,16 @@
...
+
     parser.add_argument("user", help="Identifier of the user launching this program")
     parser.add_argument("google_cloud_credentials_file_path", metavar="google-cloud-credentials-file-path",
                         help="Path to a Google Cloud service account credentials file to use to access the "
                              "credentials bucket"),
     parser.add_argument("configuration_module",
                         help="Configuration module to import e.g. 'configurations.test_config'. "
                              "This module must contain a PIPELINE_CONFIGURATION property")
+    parser.add_argument("membership_group_dir_path",
+                        help="Path to a directory to use to read & write membership group csvs to.")
```
DanielMwendwa, line 24: I suggest this should be optional and then add a condition after importing the pipeline config module to check if the   membership_group_configuration is not None, if so raise error indicating membership_group_dir_path is required.

src/engagement_db_to_analysis/analysis_files.py
```diff
@@ -98,29 +113,27 @@ def _get_analysis_file_row(column_view_td, column_configs, export_timestamps=Fal
...
     :type export_path: str
     """
     log.info(f"Exporting analysis file to '{export_path}'...")
 
-    column_configs = analysis_dataset_configs_to_column_configs(analysis_dataset_configurations)
-
     IOUtils.ensure_dirs_exist_for_file(export_path)
     with open(export_path, "w") as f:
-        headers = _get_analysis_file_headers(column_configs, export_timestamps)
+        headers = _get_analysis_file_headers(pipeline_config,  export_timestamps)
```
DanielMwendwa, line 133:
```suggestion
        headers = _get_analysis_file_headers(pipeline_config, export_timestamps)
```

src/engagement_db_to_analysis/configuration.py
```diff
@@ -44,8 +44,14 @@ class GoogleDriveUploadConfiguration:
     drive_dir: str
 
 
+@dataclass
+class MembershipGroupConfiguration(object):
```
DanielMwendwa, line 48:
```suggestion
class MembershipGroupConfiguration:
```

src/engagement_db_to_analysis/membership_group.py
```diff
@@ -0,0 +1,90 @@
+import os
+from google.api_core.exceptions import  NotFound
```
DanielMwendwa, line 2:
```suggestion
from google.api_core.exceptions import NotFound
```

src/engagement_db_to_analysis/membership_group.py
```diff
@@ -0,0 +1,90 @@
...
+def get_membership_groups_csvs(google_cloud_credentials_file_path, membership_group_csv_urls, membership_group_dir_path):
+    """
+    Downloads de-identified membership groups CSVs from g-cloud.
+    :param google_cloud_credentials_file_path: Path to the Google Cloud service account credentials file to use to
+                                               access the credentials bucket.
+    :type google_cloud_credentials_file_path: str
+    :param membership_group_csv_urls: Dict of membership group name to group g-cloud csv url(s).
+    :type membership_group_csv_urls: Dict
+    :param membership_group_dir_path: Path to directory containing de-identified membership groups CSVs containing membership groups data
+                        stored as `avf-participant-uuid` column.
```
DanielMwendwa, line 23: :type membership_group_dir_path: str

src/engagement_db_to_analysis/membership_group.py
```diff
@@ -0,0 +1,90 @@
...
+            membership_group_csv = membership_group_csv_url.split("/")[-1]
+
+            export_file_path = f'{membership_group_dir_path}/{membership_group_csv}'
+
+            if os.path.exists(export_file_path):
+                log.info(f"File '{membership_group_csv}' already exists, skipping download")
+                continue
+
+            try:
+                log.info(f"Saving '{membership_group_csv}' to file f'{membership_group_dir_path}...")
```
DanielMwendwa, line 37: `membership_group_dir_path` is a directory; you may use `export_file_path` or change the name file to dir

src/engagement_db_to_analysis/membership_group.py
```diff
@@ -0,0 +1,90 @@
...
+
+            if os.path.exists(membership_group_csv_file_path):
+                with open(membership_group_csv_file_path, "r", encoding='utf-8-sig') as f:
+                    membership_group_data = list(csv.DictReader(f))
+                    for row in membership_group_data:
+                        membership_group_participants[membership_group].add(row['avf-participant-uuid'])
+            else:
+                log.warning(f"{membership_group_csv} does not exist in {membership_group_dir_path} skipping!")
+
+        log.info(f'Loaded {len(membership_group_participants[membership_group])} {membership_group} uids')
```
DanielMwendwa, line 79: You can extract this code block into function `get_membership_groups_participants(membership_group_csv_urls, membership_group_dir_path)` so that this function only does what the function states - tagging...

src/engagement_db_to_analysis/membership_group.py
```diff
@@ -0,0 +1,90 @@
...
+def tag_membership_groups_participants(user, column_view_traced_data, membership_group_csv_urls, membership_group_dir_path):
+    """
+    This tags uids who participated in projects membership groups.
+    :param user: Identifier of the user running this program, for TracedData Metadata.
+    :type user: str
+    :param column_view_traced_data: Messages/Participants TracedData organised into column-view format.
+    :type: list of core_data_modules.traced_data.TracedData
+    :param membership_group_dir_path: Path to directory containing de-identified membership groups CSVs containing membership groups data
+                        stored as `avf-phone-uuid` and `Name` columns.
+    :type user: str
```
DanielMwendwa, line 56: replace `:type user: str` with `:type membership_group_dir_path: str`

src/engagement_db_to_analysis/membership_group.py
```diff
@@ -0,0 +1,90 @@
...
+    :param membership_group_csv_urls: Dict of membership group name to group g-cloud csv url(s).
+    :type membership_group_csv_urls: Dict
+    """
+
+    membership_group_participants = dict() # of group name to group avf-participant-uuid(s)
+
+    # Read listening group participants CSVs and add their uids to the respective group
+    for membership_group, membership_group_csv_url in membership_group_csv_urls:
+        membership_group_participants[membership_group] = set()
+        for i, membership_group_csv_url in enumerate(membership_group_csv_url):
```
DanielMwendwa, line 66: There is a slight confusion in the way `membership_group_csv_url` has been used.
I'd suggest: `for membership_group, membership_group_csv_urls in membership_group_to_csv_urls:` or `for membership_group, membership_group_csv_urls in membership_group_csv_urls_map:`
Then in line 65: `for i, membership_group_csv_url in enumerate(membership_group_csv_urls):`

src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -162,8 +164,8 @@ def export_traced_data(traced_data, export_path):
         TracedDataJsonIO.export_traced_data_iterable_to_jsonl(traced_data, f)
 
 
-def generate_analysis_files(user, google_cloud_credentials_file_path, pipeline_config, engagement_db, output_dir,
-                            cache_path=None):
+def generate_analysis_files(user, google_cloud_credentials_file_path, pipeline_config, engagement_db, membership_group_dir_path,
```
DanielMwendwa, line 167: If you'll update as suggested in engagement_db_to_analysis.py `membership_group_dir_path` will default to none.

DanielMwendwa (approved):

engagement_db_to_analysis.py
IsaackMwenda: This introduces a point of failure when we forget to pass membership_group_dir_path which is what we are trying to solve with the new pipelines. Checking for if the config exists before tagging should suffice.

src/engagement_db_to_analysis/membership_group.py
IsaackMwenda: Thanks for catching this - updated.

src/engagement_db_to_analysis/membership_group.py
IsaackMwenda: Updated for clarity - thanks

src/engagement_db_to_analysis/membership_group.py
IsaackMwenda: Reading data to tag is certainly part of the tagging process. Since we are not reusing the code block else where a function would be an overkill here.

PR #129, opened 2021-11-17
Handle empty messages

as2388:
Empty messages are defined as None or "". Handles by:
 - Not attempting to upload to Coda (Coda's api rejects these anyway)
 - Automatically labelling as NC

### DIFF
--- a/src/engagement_db_to_analysis/code_imputation_functions.py
+++ b/src/engagement_db_to_analysis/code_imputation_functions.py
@@ -222,6 +222,42 @@ def _impute_ws_coding_errors(user, messages_traced_data, analysis_dataset_config
     log.info(f"Imputed {imputed_labels} {Codes.CODING_ERROR} labels for WS codes")
 
 
+def _impute_nc_for_empty_messages(user, messages_traced_data, analysis_dataset_configs, ws_correct_dataset_code_scheme):
+    """
+    Imputes Codes.NOT_CODED for messages whose text property is either None or "".
+
+    :param user: Identifier of user running the pipeline.
+    :type user: str
+    :param messages_traced_data: Messages TracedData objects to impute age_category.
+    :type messages_traced_data: list of TracedData
+    :param analysis_dataset_configs: Analysis dataset configuration in pipeline configuration module.
+    :type analysis_dataset_configs: pipeline_config.analysis_configs.dataset_configurations
+    :param ws_correct_dataset_code_scheme: WS - Correct Dataset code scheme.
+    :type ws_correct_dataset_code_scheme: core_data_modules.data_models.CodeScheme
+    """
+    messages_with_nc_imputed = 0
+    for message_td in messages_traced_data:
+        message = Message.from_dict(dict(message_td))
+
+        if not (message.text is None or message.text == ""):
+            continue
+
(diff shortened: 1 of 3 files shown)


src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -222,6 +222,42 @@ def _impute_ws_coding_errors(user, messages_traced_data, analysis_dataset_config
...
+    :param analysis_dataset_configs: Analysis dataset configuration in pipeline configuration module.
+    :type analysis_dataset_configs: pipeline_config.analysis_configs.dataset_configurations
+    :param ws_correct_dataset_code_scheme: WS - Correct Dataset code scheme.
+    :type ws_correct_dataset_code_scheme: core_data_modules.data_models.CodeScheme
+    """
+    messages_with_nc_imputed = 0
+    for message_td in messages_traced_data:
+        message = Message.from_dict(dict(message_td))
+
+        if not (message.text is None or message.text == ""):
```
IsaackMwenda, line 242: Have we encountered this from Radpidpro before message.text is None ?

src/engagement_db_to_analysis/code_imputation_functions.py
```diff
@@ -222,6 +222,42 @@ def _impute_ws_coding_errors(user, messages_traced_data, analysis_dataset_config
...
+            continue
+
+        message_analysis_config = analysis_dataset_config_for_message(analysis_dataset_configs, message)
+        normal_code_schemes = [c.code_scheme for c in message_analysis_config.coding_configs]
+
+        _clear_latest_labels(user, message_td, normal_code_schemes + [ws_correct_dataset_code_scheme])
+        for code_scheme in normal_code_schemes:
```
IsaackMwenda, line 249: I think we want to filter them out so that we don't skew our relevance rate.

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: Yes - I checked old data from a year ago and ssf elections and saw this in both cases, but it looks like it's when the text received from Shaqodoon is "". I think we should ask (i) Shaqodoon what causes an "" to be sent, and (ii) Rapid Pro if auto-conversion to null is the right thing to do. Perhaps we can discuss tomorrow

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: NC was standard practice until now (https://github.com/AfricasVoices/Project-SSF-ELECTIONS/blob/e31b55c5e1abe191f80195f86b7a86d83e264b97/src/apply_manual_codes.py#L116-L136) - but I think now would be a good time to discuss with RDA (after understanding the causes) and see if that's still the right thing to do.

src/engagement_db_to_analysis/code_imputation_functions.py
as2388: I've checked the logs more carefully - the message from Shaqodoon contains 'text=', so I can understand why Rapid Pro interprets this as null. (It converts to "" in its acknowledgement reply to Shaqodoon, but I don't think that's a problem)

DanielMwendwa (approved):

PR #134, opened 2021-12-09
Reduce frequency of coda user ids update

DanielMwendwa:
This has been implemented on the wusc pipeline

### DIFF
--- a/src/engagement_db_coda_sync/lib.py
+++ b/src/engagement_db_coda_sync/lib.py
@@ -50,9 +50,7 @@ def ensure_coda_datasets_up_to_date(coda, coda_config, google_cloud_credentials_
             config_user_ids = default_project_user_ids
 
         coda_user_ids = coda.get_dataset_user_ids(dataset_config.coda_dataset_id)
-        if coda_user_ids is not None:
-            config_user_ids_not_in_coda = set(config_user_ids) - set(coda_user_ids)
-        if config_user_ids_not_in_coda:
+        if  set(coda_user_ids) != set(config_user_ids):
             coda.set_dataset_user_ids(dataset_config.coda_dataset_id, config_user_ids)
 
         repo_code_schemes = []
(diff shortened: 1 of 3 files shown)


src/engagement_db_coda_sync/lib.py
```diff
@@ -48,7 +48,12 @@ def ensure_coda_datasets_up_to_date(coda, coda_config, google_cloud_credentials_
             config_user_ids = get_coda_users_from_gcloud(dataset_config.dataset_users_file_url, google_cloud_credentials_file_path)
         else:
             config_user_ids = default_project_user_ids
-        coda.set_dataset_user_ids(dataset_config.coda_dataset_id, config_user_ids)
+
+        coda_user_ids = coda.get_dataset_user_ids(dataset_config.coda_dataset_id)
+        if coda_user_ids is not None:
+            config_user_ids_not_in_coda = set(config_user_ids) - set(coda_user_ids)
+        if config_user_ids_not_in_coda:
+            coda.set_dataset_user_ids(dataset_config.coda_dataset_id, config_user_ids)
```
as2388, line 56:
```suggestion
        if  set(coda_user_ids) != set(config_user_ids):
            coda.set_dataset_user_ids(dataset_config.coda_dataset_id, config_user_ids)
```
because (i) the double if, especially with the second one being outdented, is confusing, and (ii) this doesn't let us remove users from coda, because in that case set(config_user_ids) - set(coda_user_ids) can be 0 even though there are users in coda_user_ids who aren't in config_user_ids

as2388 (approved):
How necessary is this? It looks to me like we're exchanged e-6 cents of Firestore costs per run for a few additional milliseconds runtime by swapping a single write with a read and optional write.

DanielMwendwa:
I've removed this implemented on the wusc pipeline until this is resolved.

src/engagement_db_coda_sync/lib.py
DanielMwendwa: We'll still need to check for `coda_user_ids is not None`. yes, set(config_user_ids) - set(coda_user_ids) can be 0 enabling us not delete coda users. even though checking for inequality makes sense, this will let us delete users from coda which we agreed not to do instead delete manually or have a tool that does that.

So generally the main rreason for this PR is to prevent deleting users_ids from firestore.

DanielMwendwa: Got your point. Thanks.

IsaackMwenda (approved):

PR #135, opened 2021-12-14
Common get messages in datasets

as2388:
This will let us use this same fetch function to incrementally fetch messages for the rapid pro sync back in future.

### DIFF
--- a/src/common/get_messages_in_datasets.py
+++ b/src/common/get_messages_in_datasets.py
@@ -1,0 +1,110 @@
+from core_data_modules.logging import Logger
+
+log = Logger(__name__)
+
+
+def get_messages_in_datasets(engagement_db, engagement_db_datasets, cache=None):
+    """
+    Gets messages in the specified datasets.
+
+    :param engagement_db: Engagement database to fetch messages from.
+    :type engagement_db: engagement_database.EngagementDatabase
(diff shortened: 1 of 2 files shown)


src/common/get_messages_in_datasets.py
```diff
@@ -0,0 +1,110 @@
...
+            if latest_message_timestamp is not None:
+                cache.set_date_time(engagement_db_dataset, latest_message_timestamp)
+
+            if full_download_required:
+                # Export this as the ws case too, as there will be no need to check for ws messages that moved from
+                # this dataset before this initial fetch.
+                cache.set_date_time(f"{engagement_db_dataset}_ws", latest_message_timestamp)
+
+            # Export project engagement_dataset files
+            if len(messages) > 0:
```
IsaackMwenda, line 107: I think we still want to catch duplicate origin_ids here.

IsaackMwenda (approved):

DanielMwendwa (approved):

src/common/get_messages_in_datasets.py
as2388: Thanks for catching this, I think the code on main changed while I was working on this PR.
I have added this back in, and manually compared both versions to check for any other differences.

PR #136, opened 2022-01-07
Add incremental mode to the engagement db -> rapid pro sync

as2388:
Sorry for the big diff - unlike the other stages, the previous version of this script wasn't designed with a future incremental mode in mind, so took a lot more effort to make work.

### DIFF
--- a/src/engagement_db_to_rapid_pro/engagement_db_to_rapid_pro.py
+++ b/src/engagement_db_to_rapid_pro/engagement_db_to_rapid_pro.py
@@ -1,17 +1,139 @@
 import glob
 import json
+from collections import defaultdict
 
 from core_data_modules.cleaners import Codes
 from core_data_modules.data_models import CodeScheme
 from core_data_modules.logging import Logger
 from engagement_database.data_models import MessageStatuses
 
+from src.common.cache import Cache
+from src.common.get_messages_in_datasets import get_messages_in_datasets
 from src.engagement_db_to_rapid_pro.configuration import WriteModes
 
 log = Logger(__name__)
 
 # Value to write to a Rapid Pro contact field if we're only indicating the presence of an answer
 _PRESENCE_VALUE = "#ENGAGEMENT-DATABASE-HAS-RESPONSE"
+
+
+def _engagement_db_datasets_in_sync_config(sync_config):
+    """
+    :param sync_config: Sync config to get the engagement_db_datasets from.
+    :type sync_config: src.engagement_db_to_rapid_pro.configuration.EngagementDBToRapidProConfiguration
+    :return: All the engagement_db_datasets in the sync_config.
+    :rtype: set of str
+    """
+    engagement_db_datasets = set()
+
+    if sync_config.normal_datasets is not None:
+        for dataset_config in sync_config.normal_datasets:
+            engagement_db_datasets.update(dataset_config.engagement_db_datasets)
+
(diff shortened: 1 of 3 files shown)


src/engagement_db_to_rapid_pro/engagement_db_to_rapid_pro.py
```diff
@@ -14,6 +16,125 @@
...
+    all_labels = []
+    for msg in participant_messages:
+        if msg.dataset in sync_config.consent_withdrawn_dataset.engagement_db_datasets:
+            all_labels.extend(msg.get_latest_labels())
+
+    contact_fields = dict()
+    consent_withdrawn_contact_field = sync_config.consent_withdrawn_dataset.rapid_pro_contact_field
+    if _labels_contain_consent_withdrawn(all_labels, code_schemes):
+        contact_fields[consent_withdrawn_contact_field.key] = "yes"
+    elif sync_config.allow_clearing_fields:
```
IsaackMwenda, line 132: Which use case would we want to clear the consent_withdrawn_contact_field?

as2388: In cases where it had previously been set incorrectly e.g. if someone labels a message as STOP then corrects it to something else.

IsaackMwenda: Thanks makes sense.

IsaackMwenda (approved):
Looks good. Thanks @as2388.

src/engagement_db_to_rapid_pro/engagement_db_to_rapid_pro.py
```diff
@@ -88,52 +229,42 @@ def sync_engagement_db_to_rapid_pro(engagement_db, rapid_pro, uuid_table, sync_c
...
-        log.info(f"Syncing participant {i + 1}/{len(participants)}: {participant_uuid}...")
-        # Re-identify the participant
-        urn = uuid_table.uuid_to_data(participant_uuid)
+    # Sync each message to Rapid Pro, by recomputing the state of every participant.
+    participants_synced_this_cycle = set()
+    for i, message in enumerate(messages_triggering_sync):
+        log.info(f"Syncing message {i + 1}/{len(messages_triggering_sync)}: {message.message_id}...")
+        participant_uuid = message.participant_uuid
+        if participant_uuid in participants_synced_this_cycle:
+            log.info(f"Skipping this message because we've already synced participant_uuid {participant_uuid} this run")
```
DanielMwendwa, line 243:
```suggestion
            log.info(f"Skipping this message because we've already synced participant_uuid {participant_uuid} before this run")
```

DanielMwendwa (approved):

src/engagement_db_to_rapid_pro/engagement_db_to_rapid_pro.py
as2388: I think this is correct as-is because I'm using "run" to mean a complete run of this pipeline stage, and because the optimisation being made here is to fully sync each participant only once, rather than once per changed message.

Would it help to rephrase this log to we've already synced "in this pipeline run"?

DanielMwendwa: yes

PR #143, opened 2022-01-26
Add dry run to scripts involved in synchronizing Engagement db to Coda 

DanielMwendwa:


as2388 (approved):

src/engagement_db_coda_sync/engagement_db_to_coda.py
```diff
@@ -177,7 +184,7 @@ def sync_engagement_db_to_coda(engagement_db, coda, coda_config, cache_path=None
     for dataset_config in coda_config.dataset_configurations:
         log.info(f"Syncing engagement db dataset {dataset_config.engagement_db_dataset} to Coda dataset "
                  f"{dataset_config.coda_dataset_id}...")
-        dataset_sync_stats = _sync_engagement_db_dataset_to_coda(engagement_db, coda, coda_config, dataset_config, cache)
+        dataset_sync_stats = _sync_engagement_db_dataset_to_coda(engagement_db, coda, coda_config, dataset_config, cache, is_dry_run)
         dataset_to_sync_stats[dataset_config.engagement_db_dataset] = dataset_sync_stats
 
     # Log the summaries of actions taken for each dataset then for all datasets combined.
```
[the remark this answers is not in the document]
DanielMwendwa: Updated, thanks.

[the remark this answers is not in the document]
DanielMwendwa: Updated thanks.

src/engagement_db_coda_sync/engagement_db_to_coda.py
```diff
@@ -172,20 +180,26 @@ def sync_engagement_db_to_coda(engagement_db, coda, coda_config, cache_path=None
...
+
     # Log the summaries of actions taken for each dataset then for all datasets combined.
     all_sync_stats = EngagementDBToCodaSyncStats()
     for dataset_config in coda_config.dataset_configurations:
         log.info(f"Summary of actions for engagement db dataset '{dataset_config.engagement_db_dataset}':")
         dataset_to_sync_stats[dataset_config.engagement_db_dataset].print_summary()
         all_sync_stats.add_stats(dataset_to_sync_stats[dataset_config.engagement_db_dataset])
 
-    log.info(f"Summary of actions for all datasets:")
+    log.info(f"Summary of actions for all datasets: {dry_run_text}")
```
IsaackMwenda, line 204:
```suggestion
    log.info(f"Summary of actions for all datasets with dry run set to : {dry_run}")
```
Would be more expressive

sync_engagement_db_to_coda.py
```diff
@@ -35,6 +39,9 @@
     project = subprocess.check_output(["git", "config", "--get", "remote.origin.url"]).decode().strip()
 
     HistoryEntryOrigin.set_defaults(user, project, pipeline, commit)
+    
+    dry_run_text = "(dry run)" if dry_run else ""
+    log.info(f"Synchronizing data from an engagement database to Coda {dry_run_text}")
```
IsaackMwenda, line 44: ditto

IsaackMwenda (approved):
Thanks.

src/engagement_db_coda_sync/engagement_db_to_coda.py
DanielMwendwa: I think we should not change that because of consistency with other scripts i.e., restore_engagement_db in Tools repo and also it will be easier to figure if a run is in dry or not if you spot `dry run` in the logs instead of checking if it has been set to True or False.

IsaackMwenda: I still think printing the bool for a dry run gives more information, e.g when compared to printing empty string when the dry run is False. If you agree let's update restore_engagement_db to match the new log format in a diff PR. 

(am not strongly inclined to any of them - pick any that you think is more descriptive)

as2388: I agree with @DanielMwendwa - if you haven't opted into dry-run then you shouldn't need to see it in the logs, and the expectation should be if the log says something has been done, then that's what was done unless it caveats that with the dry_run note.

(The compromise option would be to print (dry run) or something like (production run/non dry run) instead of True/False)

PR #144, opened 2022-02-04
Update cache incrementally in coda -> db sync

as2388:


DanielMwendwa (approved):

src/engagement_db_coda_sync/coda_to_engagement_db.py
```diff
@@ -88,10 +90,10 @@ def _sync_coda_dataset_to_engagement_db(coda, engagement_db, coda_config, datase
         )
         sync_stats.add_stats(message_sync_stats)
 
-    seen_timestamps = [msg.last_updated for msg in coda_messages if msg.last_updated is not None]
-    if cache is not None and len(seen_timestamps) > 0:
-        most_recently_updated_timestamp = sorted(seen_timestamps)[-1]
-        cache.set_last_updated_timestamp(dataset_config.coda_dataset_id, most_recently_updated_timestamp)
+        # If there's a cache and we've read the last message, or the next message has a different modification
```
IsaackMwenda, line 93:
```suggestion
        # If there's a cache and we've read the last message, or the next message last updated timestamp is greater than the message we are currently syncing, update the cache.
```

IsaackMwenda (approved):

src/engagement_db_coda_sync/coda_to_engagement_db.py
as2388: Thanks, this is much clearer so I've changed it to this.

PR #145, opened 2022-02-04
Batch engagement db reads in coda -> db sync

as2388:
This fixes the problem of pipelines crashing when a Coda message has more than 250 matching engagement db messages

### DIFF
--- a/src/engagement_db_coda_sync/coda_to_engagement_db.py
+++ b/src/engagement_db_coda_sync/coda_to_engagement_db.py
@@ -10,10 +10,10 @@ log = Logger(__name__)
 
 
 @firestore.transactional
-def _sync_coda_message_to_engagement_db(transaction, coda_message, engagement_db, engagement_db_dataset, coda_config):
+def _sync_coda_message_to_engagement_db_batch(transaction, coda_message, engagement_db, engagement_db_dataset,
+                                              coda_config, start_after=None):
     """
-    Syncs a coda message to an engagement database, by downloading all the engagement database messages which match the
-    coda message's id and dataset, and making sure the labels match.
+    Syncs a Coda message to a batch of up to 250 engagement database messages.
 
     :param transaction: Transaction in the engagement database to perform the update in.
     :type transaction: google.cloud.firestore.Transaction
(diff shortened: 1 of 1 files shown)


src/engagement_db_coda_sync/coda_to_engagement_db.py
```diff
@@ -10,10 +10,10 @@
 
 
 @firestore.transactional
-def _sync_coda_message_to_engagement_db(transaction, coda_message, engagement_db, engagement_db_dataset, coda_config):
+def _sync_coda_message_to_engagement_db_batch(transaction, coda_message, engagement_db, engagement_db_dataset,
+                                              coda_config, start_after=None):
     """
-    Syncs a coda message to an engagement database, by downloading all the engagement database messages which match the
-    coda message's id and dataset, and making sure the labels match.
+    Syncs a Coda message to a batch of up to 250 engagement database messages.
```
IsaackMwenda, line 16: The firestore batched write limit is 500 whats the reason for using 250?

IsaackMwenda (approved):
LGTM with a comment thanks.

src/engagement_db_coda_sync/coda_to_engagement_db.py
as2388: We need one read and one write per message, so it's 250 x 2 = 500.
I can add a comment that clarifies in the code

DanielMwendwa (approved):

src/engagement_db_coda_sync/coda_to_engagement_db.py
DanielMwendwa: I think this is caused by a different issue. Check this [firebase issue](https://github.com/firebase/firebase-admin-node/issues/456#issuecomment-464187898). Maybe we should try writing without including server timestamps to confirm.

as2388: We're both wrong here. I've done some tests to verify the behaviour and find that:
 - Reads don't ever affect batches, so it's definitely not this.
 - The server timestamp double write cost issue was fixed a few months ago (I remember reading about this policy changing at the time, but am now struggling to find it in the Firestore release notes. I have just written 500 documents with server timestamps in a single batch though, so I could at least verify this double-write cost no longer applies despite struggling to find an official record of the change just now)

The actual reason is that we decided every write to the database would get an associated history_entry written too - so it's 2 writes per doc, 1 to update the message and 1 to write the history record.

PR #150, opened 2022-02-18
Db -> Analysis dry run

DanielMwendwa:


src/engagement_db_to_analysis/engagement_db_to_analysis.py
```diff
@@ -111,6 +111,8 @@ def generate_analysis_files(user, google_cloud_credentials_file_path, pipeline_c
     if pipeline_config.analysis.google_drive_upload is None:
         log.debug("Not uploading to Google Drive, because the 'google_drive_upload' configuration was None")
     else:
+        if dry_run:
+            exit(0)
```
as2388, line 115: Prefer
```suggestion
    if pipeline_config.analysis.google_drive_upload is None or dry_run:
        log.debug("Not uploading to Google Drive, because the 'google_drive_upload' configuration was None")
    else:
```
so we don't have to think about possible early exits if we add any code after this block.

engagement_db_to_analysis.py
```diff
@@ -10,25 +10,29 @@
...
+    parser.add_argument("--dry-run", action="store_true",
+                        help="Logs the updates that would be made without updating anything.")
     parser.add_argument("--incremental-cache-path",
                         help="Path to a directory to use to cache results needed for incremental operation.")
 
     parser.add_argument("user", help="Identifier of the user launching this program")
     parser.add_argument("google_cloud_credentials_file_path", metavar="google-cloud-credentials-file-path",
                         help="Path to a Google Cloud service account credentials file to use to access the "
-                             "credentials bucket"),
+                        "credentials bucket"),
```
as2388, line 21: (did you mean to reformat this?)

as2388 (approved):
Should we make sure to not update the cache on a dry-run?

IsaackMwenda (approved):

IsaackMwenda:
> Should we make sure to not update the cache on a dry-run?

we should not be updating the cache in dry runs as this will affect the next normal run - or probably have a different cache volume for dry runs.

engagement_db_to_analysis.py
DanielMwendwa: No, updated.

DanielMwendwa:
> update the cache on a dry-run

Skipped updating cache during run as done in other PRs

PR #151, opened 2022-02-21
Add csv -> engagement db sync

as2388:
This lets us sync de-identified CSVs of the type we receive from Shaqodoon to the engagement db. --dry-run, dockerise, and sync stats to come in future PRs. 

A couple of points to draw special attention to in case these should be changed:
- This uses csv hash/row indexes in the origin id, which means that csvs can't be changed after a sync without causing messages to be duplicated. The alternative would have been to look at the message text and timestamp pair, but that doesn't allow for genuine duplicated messages (recovery csvs usually have pretty poor timestamp resolution, rounded to the nearest minute).
- It only allows each csv to be pointed to one dataset. We could add timestamp filters in future if we need it, or simply break the csvs by week when we prepare them.

### DIFF
--- a/src/csv_to_engagement_db/csv_to_engagement_db.py
+++ b/src/csv_to_engagement_db/csv_to_engagement_db.py
@@ -58,7 +58,7 @@ def _csv_message_to_engagement_db_message(csv_message, uuid_table, dataset, orig
     channel_operator = URNCleaner.clean_operator(participant_urn)
 
     return Message(
-        participant_uuid=csv_message["Sender"],
+        participant_uuid=participant_uuid,
         text=csv_message["Message"],
         timestamp=_parse_date_string(csv_message["ReceivedOn"], timezone),
         direction=MessageDirections.IN,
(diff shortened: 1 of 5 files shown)


src/csv_to_engagement_db/csv_to_engagement_db.py
```diff
@@ -0,0 +1,175 @@
...
+    :rtype: engagement_database.data_models.Message
+    """
+    participant_uuid = csv_message["Sender"]
+    assert participant_uuid.startswith(uuid_table._uuid_prefix), f"Sender uuid does not start with uuid prefix " \
+                                                                 f"'{uuid_table._uuid_prefix}'"
+    participant_urn = uuid_table.uuid_to_data(participant_uuid)
+    channel_operator = URNCleaner.clean_operator(participant_urn)
+
+    return Message(
+        participant_uuid=csv_message["Sender"],
```
IsaackMwenda, line 61:
```suggestion
        participant_uuid=participant_uuid,
```

IsaackMwenda (approved):
lgtm thanks @as2388

DanielMwendwa (approved):

PR #153, opened 2022-02-24
Dockerise upload archive files

as2388:


DanielMwendwa (approved):

IsaackMwenda (commented):
I think the docker caveat for this script was we were not able to delete old files on local disk - have you tested that works?

as2388:
> I think the docker caveat for this script was we were not able to delete old files on local disk - have you tested that works?

Yes I've tested, both locally and by setting up and running the test pipeline on Miranda. Previously we couldn't run in docker because archives were incompatible with volumes and docker cp, so this uses the newer, simplified interface for bind mounts instead.

IsaackMwenda:
> > I think the docker caveat for this script was we were not able to delete old files on local disk - have you tested that works?
> 
> Yes I've tested, both locally and by setting up and running the test pipeline on Miranda. Previously we couldn't run in docker because archives were incompatible with volumes and docker cp, so this uses the newer, simplified interface for bind mounts instead.

Great thanks.

IsaackMwenda (approved):

PR #155, opened 2022-02-25
Support multiple datasets per csv file, filtering by message timestamp

as2388:
In #151 I suggested timestamp filters weren't a priority, especially as we could create one csv per dataset. However, having tried to use that configuration on the imaqal datasets, I've learned not having timestamp filtering makes things really tedious, and it makes adjusting dates really difficult. This makes configuring csvs earlier by allowing a single csv to point to multiple datasets.

### DIFF
--- a/src/csv_to_engagement_db/configuration.py
+++ b/src/csv_to_engagement_db/configuration.py
@@ -1,5 +1,4 @@
 from datetime import datetime
-
 import pytz
 
 
--- a/src/csv_to_engagement_db/configuration.py
+++ b/src/csv_to_engagement_db/configuration.py
@@ -1,12 +1,88 @@
-class CSVSource:
-    def __init__(self, gs_url, engagement_db_dataset, timezone):
-        self.gs_url = gs_url
+from datetime import datetime
+
+import pytz
+
+
+_MIN_DATE_UTC = pytz.timezone("utc").localize(datetime.min)
+_MAX_DATE_UTC = pytz.timezone("utc").localize(datetime.max)
+
+
+class CSVDatasetConfiguration:
+    def __init__(self, engagement_db_dataset, start_date=_MIN_DATE_UTC, end_date=_MAX_DATE_UTC):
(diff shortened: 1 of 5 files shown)


src/csv_to_engagement_db/configuration.py
```diff
@@ -1,12 +1,88 @@
...
+        self.start_date = start_date
+        self.end_date = end_date
+
+    def to_dict(self):
+        return {
+            "engagement_db_dataset": self.engagement_db_dataset,
+            "start_date": self.start_date,
+            "end_date": self.end_date
+        }
+
```
IsaackMwenda, line 35:
```suggestion
```

src/csv_to_engagement_db/configuration.py
```diff
@@ -1,12 +1,88 @@
+from datetime import datetime
+
```
IsaackMwenda, line 2:
```suggestion
```

IsaackMwenda (approved):
LGTM thanks.

src/csv_to_engagement_db/configuration.py
as2388: I've removed this but note that pycharm's optimise imports inserts newlines between different types of import automatically, so we'd need to add an exception to the auto-formatting rules and run a global optimise imports if we wanted to do this everywhere

src/csv_to_engagement_db/configuration.py
as2388: I would prefer not to remove this line as it would break the formatting standard that the python community has widely adopted https://www.python.org/dev/peps/pep-0008/#blank-lines

DanielMwendwa (approved):

src/csv_to_engagement_db/configuration.py
IsaackMwenda: Sorry on this - I saw 3 lines here instead of 2. It's okay as is.

PR #156, opened 2022-02-28
Facebook to db sync

IsaackMwenda:
This adds support to download and sync Facebook comments to db. Upcoming PRs
1. Handle analysis for Facebook data.
2. Facebook metrics
3. Enable caching & docker.
4. Sync stats.


as2388 (approved):
This looks great

src/social_media_to_engagement_db/facebook_to_engagement_db.py
```diff
@@ -0,0 +1,194 @@
...
+                    comment["parent"] = {}
+
+                origin_id = f'page_id_{facebook_source.page_id}.user_id_{comment["from"]["id"]}._comment_id_{comment["id"]}'
+                message = _facebook_comment_to_engagement_db_message(comment, dataset.engagement_db_dataset,
+                                                                     origin_id, uuid_table)
+
+                message_origin_details = {
+                    "page_id": facebook_source.page_id,
+                    "post_id": post_id,
+                    "user_id": comment["from"]["id"],
```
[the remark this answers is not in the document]
IsaackMwenda: That's right updated.

src/facebook_to_engagement_db/configuration.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class FacebookDataset:
+    def __init__(self, engagement_db_dataset, post_ids=None, search=None):
+        self.engagement_db_dataset = engagement_db_dataset
+        self.post_ids = post_ids
+        self.search = search
+
+        if search is not None:
+            self.search = FacebookSearch.to_dict(search)
```
DanielMwendwa, line 25: I think this should be in the `to_dict` function

src/facebook_to_engagement_db/configuration.py
```diff
@@ -0,0 +1,49 @@
...
+
+        assert self.post_ids is not None or self.search is not None, \
+            "Must provide at least a post_id or search"
```
DanielMwendwa, line 28: you can remove the default arguments for this parameters, so that we don't require this assertion.

DanielMwendwa (approved):

src/facebook_to_engagement_db/configuration.py
IsaackMwenda: This is cleaner to avoid abstracting the to_dict function.

src/facebook_to_engagement_db/configuration.py
IsaackMwenda: We need either or both - default is therefore necessary to act as a placeholder.

PR #157, opened 2022-03-01
Optimise fetching fields from the same flows

DanielMwendwa:
#### I've tested (dry run) the code on the test config. I'll continue doing more tests. Below are some of the key changes made:
Group flow configs by flow name
Download runs once for each flow
Update cache by flow id without specifying rapidpro result field
Updated `dataset_to_sync_stats` generation because we are now downloading runs once and also using events associated with this runs to generate stats for every dataset we are syncing.


src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -174,15 +172,20 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
...
+    groups = []
+    unique_keys = []
+    flow_result_configurations = sorted(rapid_pro_config.flow_result_configurations, key=lambda x: x.flow_name)
+    for flow_name, flow_configs in groupby(flow_result_configurations, lambda x: x.flow_name):
+        groups.append(list(flow_configs))
+        unique_keys.append(flow_name)
+
+    flow_stats = dict() # of flow_name -> FlowStats
+    dataset_to_sync_stats = defaultdict(lambda: FlowResultToEngagementDBSyncStats())  # of '{flow_name}.{flow_result_field}' -> FlowResultToEngagementDBSyncStats
+    for flow_name, flow_configs in zip(unique_keys, groups):
```
as2388, line 184: Instead of creating 2 lists then zipping them together, I think you could use a dictionary instead. That way there's no risk of the lists ever getting out of sync with each other before you zip

a rough sketch of this would look like:
```python
flow_name_to_flow_configs = dict()
for flow_result_config in flow_result_configurations:
    flow_name_to_flow_configs[flow_result_config.flow_name] = flow_result_config

for flow_name, flow_configs in flow_name_to_flow_configs.items():
    ...
```

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -174,15 +172,20 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
...
+        stats = FlowStats()
```
as2388, line 185:
```suggestion
        flow_stats = FlowStats()
```
to distinguish from the other type of stats we're using in this stage

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -191,30 +194,28 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
             cache.set_contacts(contacts)
         contacts_lut = {c.uuid: c for c in contacts}
 
-        # Process each run in turn, adding it the engagement database if it contains a message relevant to this flow
-        # config and the message hasn't already been added to the engagement database.
-        log.info(f"Processing {len(runs)} new runs for flow '{flow_config.flow_name}'")
+        # Process each run in turn, adding its values to the engagement database if it contains messages relevant to this flow
```
as2388, line 197:
```suggestion
        # Process each run in turn, adding its values to the engagement database if it contains messages relevant to these flow
```

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -226,60 +227,82 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
...
+                        "flow_name": flow_name,
+                        "run_value": rapid_pro_result.serialize()
+                    }
+                    sync_event = _ensure_engagement_db_has_message(engagement_db, msg, message_origin_details, dry_run)
+                    sync_stats.add_event(sync_event)
+                dataset_to_sync_stats[f"{flow_name}.{config.flow_result_field}"].add_stats(sync_stats)
+            
             # Update the cache so we know not to check this run again in this flow + result field context.
+            # TODO: Update the cache if we've read the last run, or the next run's last modified timestamp is greater
+            # than the run we are currently syncing.
```
as2388, line 281: This seems like quite an important todo which would should be addressed now

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -226,60 +227,82 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
...
 
-    # Log the summaries of actions taken for each dataset then for all datasets combined.
-    all_sync_stats = RapidProToEngagementDBSyncStats()
-    for flow_config in rapid_pro_config.flow_result_configurations:
-        result_field = f"{flow_config.flow_name}.{flow_config.flow_result_field}"
-        log.info(f"Summary of actions for flow result field '{result_field}':")
-        dataset_to_sync_stats[result_field].print_summary()
-        all_sync_stats.add_stats(dataset_to_sync_stats[result_field])
+        for _ in runs:
+            stats.add_event(RapidProSyncEvents.READ_RUN_FROM_RAPID_PRO)
```
as2388, line 286: Can you move this to the top of the runs loop and add the events incrementally?

src/rapid_pro_to_engagement_db/sync_stats.py
```diff
@@ -13,25 +13,37 @@ class RapidProSyncEvents:
...
             RapidProSyncEvents.RUN_CONTACT_UUID_NOT_IN_CONTACTS: 0,
             RapidProSyncEvents.UUID_FILTER_CONTACT_NOT_IN_UUID_TABLE: 0,
             RapidProSyncEvents.CONTACT_NOT_IN_UUID_FILTER: 0,
-            RapidProSyncEvents.MESSAGE_ALREADY_IN_ENGAGEMENT_DB: 0,
-            RapidProSyncEvents.ADD_MESSAGE_TO_ENGAGEMENT_DB: 0
         })
 
     def print_summary(self):
         log.info(f"Runs downloaded from Rapid Pro: {self.event_counts[RapidProSyncEvents.READ_RUN_FROM_RAPID_PRO]}")
+        log.info("The runs specified below were not processed:")
```
as2388, line 31: I'm not convinced this is necessary. But if you think it helps please can you at least rephrase it so it's clear totals are expected. "runs specified below" implies that some run ids are about to be logged, which isn't the case.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -226,60 +227,82 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
...
+        flow_stats[flow_name] = stats
+
+    # Log the summaries of actions taken for each flow and each dataset then for all flows and datasets combined.
+    all_flow_stats = FlowStats()
+    all_sync_stats = FlowResultToEngagementDBSyncStats()
```
as2388, line 291: This is a really nice solution - we should consider something similar in the CSV sync

as2388 (changes requested):
For the most part this looks great.

My main concern is the approach to caching - grouping flow results like this means if we change which result fields we're interested in for a particular flow, then we need to reprocess all the runs. At the moment, you have to manually remember to check whether the flow configs have changed and then manually reset the cache, which I think is too risky. Maybe have something which checks the configs are the same before proceeding with cached data?

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
DanielMwendwa: Thanks. I'll work on it.

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
DanielMwendwa: Updated. Thanks.

DanielMwendwa:
> For the most part this looks great.
> 
> My main concern is the approach to caching - grouping flow results like this means if we change which result fields we're interested in for a particular flow, then we need to reprocess all the runs. At the moment, you have to manually remember to check whether the flow configs have changed and then manually reset the cache, which I think is too risky. Maybe have something which checks the configs are the same before proceeding with cached data?

Thanks, I've added a section that checks the configs are the same before proceeding with cached data - it saves the flow configs in the cache every time there is a change in the configs.

src/common/cache.py
```diff
@@ -90,3 +90,8 @@ def set_messages(self, entry_name, messages):
         with open(export_file_path, "w") as f:
             for msg in messages:
                 f.write(f"{json.dumps(msg.to_dict(serialize_datetimes_to_str=True))}\n")
+
+    def delete_file(self, entry_name):
+        file_path = f"{self.cache_dir}/{entry_name}.txt"
+        IOUtils.ensure_dirs_exist_for_file(file_path)
+        remove(file_path)
```
as2388, line 97: Not all cache files are text files, so I think this should be def delete_file(self, filename)

src/common/cache.py
```diff
@@ -90,3 +90,8 @@ def set_messages(self, entry_name, messages):
         with open(export_file_path, "w") as f:
             for msg in messages:
                 f.write(f"{json.dumps(msg.to_dict(serialize_datetimes_to_str=True))}\n")
+
+    def delete_file(self, entry_name):
```
as2388, line 94: I also think use of the file system is arguably implementation specific, so suggest making delete_file private and adding clear_timestamp(self, entry_name) too.

as2388, line 94: I also think use of the file system is arguably implementation specific, so suggest making delete_file private and adding clear_timestamp(self, entry_name) too.

src/rapid_pro_to_engagement_db/cache.py
```diff
@@ -20,28 +24,47 @@ def set_contacts(self, contacts):
...
+
+    def set_flow_result_configs(self, configs):
+        export_path = f"{self.cache_dir}/flow_result_configurations.json"
+        IOUtils.ensure_dirs_exist_for_file(export_path)
+        with open(export_path, "w") as f:
+            json.dump([c.to_dict() for c in configs], f)
+
+    def get_flow_result_configs(self):
+        try:
+            print(f"{self.cache_dir}/flow_result_configurations.json")
```
as2388, line 66: log with a message (or remove)

as2388, line 66: log with a message (or remove)

src/rapid_pro_to_engagement_db/configuration.py
```diff
@@ -1,20 +1,39 @@
...
+from typing import Dict, Optional
 
 
-@dataclass
 class FlowResultConfiguration:
-    flow_name: str
-    flow_result_field: str
-    engagement_db_dataset: str
+
+    def __init__(self, flow_name: str, flow_result_field: str, engagement_db_dataset: str):
```
as2388, line 8: Can you add a todo to add docstrings for these so they follow the format for the rest of our classes?

as2388, line 8: Can you add a todo to add docstrings for these so they follow the format for the rest of our classes?

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -174,15 +180,34 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
...
+        updated_flow_result_configs = [config for config in rapid_pro_config.flow_result_configurations \
+                                        if config.to_dict() not in cached_flow_result_configs]            
+        if len(updated_flow_result_configs) > 0:
+            seen = set()
+            for config in updated_flow_result_configs:
+                if config.flow_name not in seen:
+                    cache.reset_latest_run_timestamp(rapid_pro.get_flow_id(config.flow_name))
+                    seen.add(config.flow_name)
+            if not dry_run:
+                cache.set_flow_result_configs(rapid_pro_config.flow_result_configurations)
```
as2388, line 198: Can you extract this into a function?

as2388, line 198: Can you extract this into a function?

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
```diff
@@ -174,15 +180,34 @@ def sync_rapid_pro_to_engagement_db(rapid_pro, engagement_db, uuid_table, rapid_
...
-        sync_stats = RapidProToEngagementDBSyncStats()
+    # Check the configs are the same before proceeding with cached data
+    cached_flow_result_configs = _get_flow_result_configs_from_cache(cache)
+    if cached_flow_result_configs is None and cache is not None and not dry_run:
+        cache.set_flow_result_configs(rapid_pro_config.flow_result_configurations)
+    else:
+        cached_flow_result_configs = [d.to_dict() for d in cached_flow_result_configs]
+        updated_flow_result_configs = [config for config in rapid_pro_config.flow_result_configurations \
+                                        if config.to_dict() not in cached_flow_result_configs]            
+        if len(updated_flow_result_configs) > 0:
```
as2388, line 191: This covers the case I was worried about, which is that if you added a new row to the flow configs it wouldn't be obvious from the structure of the flow configurations that a cache reset was needed. Can you add a todo for the case it doesn't yet cover, which is if you remove a flow configuration, re-run, then add the same configuration back again, which might happen if you were switching between branches.

as2388, line 191: This covers the case I was worried about, which is that if you added a new row to the flow configs it wouldn't be obvious from the structure of the flow configurations that a cache reset was needed. Can you add a todo for the case it doesn't yet cover, which is if you remove a flow configuration, re-run, then add the same configuration back again, which might happen if you were switching between branches.

as2388 (approved):

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
DanielMwendwa: extracted into `_update_cache_with_changes_in_flow_result_configs`

src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py
DanielMwendwa: Added in L52-53

src/rapid_pro_to_engagement_db/configuration.py
DanielMwendwa: Added

src/rapid_pro_to_engagement_db/cache.py
DanielMwendwa: removed.

src/common/cache.py
DanielMwendwa: updated

PR #163, opened 2022-03-09
Add google cloud storage dependency

DanielMwendwa:


IsaackMwenda (approved):

as2388:
What was the error that means you need to add this here? I thought all our calls to gcs were wrapped by pipeline-infrastructure, so in theory we shouldn't need that dependency here.

DanielMwendwa:
> What was the error that means you need to add this here? I thought all our calls to gcs were wrapped by pipeline-infrastructure, so in theory we shouldn't need that dependency here.

The error was `ModuleNotFoundError: No module named 'storage'` and it appeared before we dockerized the script that uploads archive files. I'll close the PR since everything works as intended after that update. Thanks.

PR #165, opened 2022-03-11
Add cache to fb engagement db

IsaackMwenda:
Facebook post endpoint does not have time parameters so we can't really benefit from our usual cache logic where we only download comments created after our last seen timestamp. This adds logic to minimize the number of reads to the db when checking if the comment already exists by syncing only comments created after our last run.


src/facebook_to_engagement_db/facebook_to_engagement_db.py
```diff
@@ -142,9 +148,11 @@ def _fetch_and_sync_facebook_to_engagement_db(google_cloud_credentials_file_path
     for dataset in facebook_source.datasets:
         # Download and sync all the comments on all the posts in this dataset.
         for post_id in _get_facebook_post_ids(facebook, facebook_source.page_id, search=dataset.search):
+            latest_comment_timestamp = None if cache is None else cache.get_latest_comment_timestamp(post_id)
             post_comments = facebook.get_all_comments_on_post(post_id,
-                        fields=["from{id}", "parent", "attachments", "created_time", "message"]
-                    )
+                                                              fields=["from{id}", "parent", "attachments",
+                                                                      "created_time", "message"],
+                                                              )
```
as2388, line 155: Is there an auto-format rule somewhere that is causing changes like this?

In general, I think it is better to indent args like this by 1 tab rather than at the brackets, in order to stop all the arguments getting crammed in the right edge of the screen
e.g. 
```
post_comments = get_all_comment_on_post(
    post_id, ["parent", "attachments", "created_time", "message"]
)
```
is slightly more easily read than:
```
post_comments = get_all_comments_on_post(post_id,
                                         ["parent",
                                          "attachments",
                                          "created_time",
                                         "message"]
                                         )
```

src/facebook_to_engagement_db/facebook_to_engagement_db.py
```diff
@@ -159,21 +167,34 @@ def _fetch_and_sync_facebook_to_engagement_db(google_cloud_credentials_file_path
                 if "parent" not in comment:
                     comment["parent"] = {}
 
-                origin_id = f'facebook_comment_id_{comment["id"]}'
-                message = _facebook_comment_to_engagement_db_message(comment, dataset.engagement_db_dataset,
-                                                                     origin_id, uuid_table)
+                #Only try to add the db comments that were created after the last seen comment.created_time
+                #This helps us reduce the number of reads to the db when checking for existing comments.
```
as2388, line 171:
```suggestion
                # Only try to add the db comments that were created after the last seen comment.created_time
                # This helps us reduce the number of reads to the db when checking for existing comments.
```

src/facebook_to_engagement_db/facebook_to_engagement_db.py
```diff
@@ -159,21 +167,34 @@ def _fetch_and_sync_facebook_to_engagement_db(google_cloud_credentials_file_path
...
+                        "comment_id": comment["id"],
+                    }
 
-                _ensure_engagement_db_has_comment(engagement_db, message, message_origin_details)
+                    _ensure_engagement_db_has_comment(engagement_db, message, message_origin_details)
 
+                    if cache is not None:
+                        cache.set_latest_comment_timestamp(post_id, isoparse(comment['created_time']))
+                else:
+                    log.info(f'Comment synced in previous run skipping ...')
```
as2388, line 193: Can you tweak this to be
```
if not add_comment_db:
    continue

# Add the comment to the db
origin_id = ...
...
```

In otherwords, keeping the mainline logic unindented

as2388 (approved):

PR #171, opened 2022-03-15
Dockerise facebook engagement db

IsaackMwenda:


DanielMwendwa (approved):

docker-run-facebook-to-engagement-db.sh
```diff
@@ -0,0 +1,72 @@
...
+        *)
+            break;;
+    esac
+done
+
+# Check that the correct number of arguments were provided.
+if [[ $# -ne 4 ]]; then
+    echo "Usage: $0 
+    [--incremental-cache-volume <incremental-cache-volume>] 
+    <user> <google-cloud-credentials-file-path> <configuration-module> <data-dir>"
```
as2388, line 26: Can you align these args with the recently merged https://github.com/AfricasVoices/Test-Pipeline-Engagement-DB/pull/159? i.e. copy in configuration and code schemes rather than requiring them at build time

IsaackMwenda: updated thanks.

PR #173, opened 2022-03-16
Ensure the correct dataset != the message's current dataset

as2388:
This lets us find labelling errors earlier, and prevents the duplicated entry in previous_datasets (for example, under the previous code, if a message was, say, in a location dataset and labelled as WS -> location, it could be redirected to location once before the problem was detected. Then, after fixing the label, be redirected to age, leaving the prev_datasets as ["location", "location"], which isn't problematic, but does look slightly confusing)

### DIFF
--- a/src/engagement_db_coda_sync/lib.py
+++ b/src/engagement_db_coda_sync/lib.py
@@ -298,6 +298,7 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
             correct_dataset = coda_config.default_ws_dataset
 
         # Ensure the message isn't being WS-corrected to the dataset it's already in.
+        # TODO: Handle this case without crashing.
         assert correct_dataset != engagement_db_message.dataset, \
             f"Engagement db message '{engagement_db_message.message_id}' (text '{engagement_db_message.text}') " \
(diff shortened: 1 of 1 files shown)


src/engagement_db_coda_sync/lib.py
```diff
@@ -297,6 +297,11 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
                 raise e
             correct_dataset = coda_config.default_ws_dataset
 
+        # Ensure the message isn't being WS-corrected to the dataset it's already in.
+        assert correct_dataset != engagement_db_message.dataset, \
+            f"Engagement db message '{engagement_db_message.message_id}' (text '{engagement_db_message.text}') " \
+            f"is being WS-corrected to dataset '{correct_dataset}', but is currently in this dataset already."
+
```
IsaackMwenda, line 304: add a todo to add a solution to the issue once we have studied it.

IsaackMwenda (approved):

DanielMwendwa (approved):

src/engagement_db_coda_sync/lib.py
as2388: agreed, added

PR #176, opened 2022-03-17
Optionally set the correct_dataset from ws_code.string_value

as2388:
Copied from https://github.com/AfricasVoices/Project-RVI-Elections/pull/9/files, but put behind an optional flag this time so we don't use this behaviour by accident.

### DIFF
--- a/src/engagement_db_coda_sync/lib.py
+++ b/src/engagement_db_coda_sync/lib.py
@@ -288,11 +288,12 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
     # WS-correct if there is a valid ws_code
     if ws_code is not None:
         # Establish the correct dataset to move this message to.
-        # To determine the dataset, the following strategies are tried, in order:
-        #  1. Search the other dataset configurations for a match.
-        #  2. If `set_dataset_from_ws_string_value` has been set, move the message to the dataset `ws_code.string_value`
+        # To determine the dataset, the following strategies are tried, in this order:
+        #  1. Search the other dataset configurations for a match. If there is no match:
+        #  2. If `set_dataset_from_ws_string_value` has been set, move the message to the dataset
+        #     `ws_code.string_value`. Otherwise:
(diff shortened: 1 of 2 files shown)


src/engagement_db_coda_sync/lib.py
```diff
@@ -287,15 +287,25 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
 
     # WS-correct if there is a valid ws_code
     if ws_code is not None:
+        # Establish the correct dataset to move this message to.
+        # To determine the dataset, the following strategies are tried, in order:
+        #  1. Search the other dataset configurations for a match.
+        #  2. If `set_dataset_from_ws_string_value` has been set, move the message to the dataset `ws_code.string_value`
+        #  3. If the `default_ws_dataset` has been specified, move the message to this default dataset.
+        #  4. Raise a ValueError.
```
DanielMwendwa, line 295: Should there be an else clause like - "in case we can't find the correct dataset from the config then...`

DanielMwendwa (approved):

src/engagement_db_coda_sync/lib.py
```diff
@@ -287,15 +287,25 @@ def _update_engagement_db_message_from_coda_message(engagement_db, engagement_db
...
         try:
             correct_dataset = \
                 coda_config.get_dataset_config_by_ws_code_string_value(ws_code.string_value).engagement_db_dataset
```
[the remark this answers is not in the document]
as2388: yes, let me do in a separate PR

src/engagement_db_coda_sync/lib.py
as2388: 4. was implicitly the else clause - but I'll edit to make it more obvious.

as2388: I've adjusted this to try to make it clearer

IsaackMwenda (approved):

PR #179, opened 2022-03-22
Add fb sync stats

IsaackMwenda:


DanielMwendwa (approved):

src/facebook_to_engagement_db/facebook_to_engagement_db.py
```diff
@@ -68,7 +68,7 @@ def _engagement_db_has_message(engagement_db, message):
     return len(matching_messages) > 0
 
 
-def _ensure_engagement_db_has_comment(engagement_db, facebook_comment, message_origin_details):
+def _ensure_engagement_db_has_comment(engagement_db, facebook_comment, message_origin_details, sync_stats):
```
as2388, line 71: As discussed offline, I think we should pass stats out via return rather than updating this reference.
e.g. https://github.com/AfricasVoices/Engagement-Data-Pipeline/blob/89ba75d56c572a35192dd84c9f18aaca1baa6ea9/src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py#L103

as2388, line 71: As discussed offline, I think we should pass stats out via return rather than updating this reference.
e.g. https://github.com/AfricasVoices/Engagement-Data-Pipeline/blob/89ba75d56c572a35192dd84c9f18aaca1baa6ea9/src/rapid_pro_to_engagement_db/rapid_pro_to_engagement_db.py#L103

as2388 (approved):

PR #182, opened 2022-04-04
Sync advert contacts to rapidpro

IsaackMwenda:
Copied and modified to fit all projects from https://github.com/AfricasVoices/Project-WUSC-LEAP-II.

### DIFF
--- a/src/engagement_db_to_analysis/rapid_pro_advert_functions.py
+++ b/src/engagement_db_to_analysis/rapid_pro_advert_functions.py
@@ -187,7 +187,7 @@ def _sync_advert_contacts_fields_to_rapid_pro(cache, target_uuids, advert_contac
         log.info("Found 0 uuids to sync in this run skipping...")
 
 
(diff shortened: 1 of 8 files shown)


src/engagement_db_to_analysis/rapid_pro_advert_functions.py
```diff
@@ -0,0 +1,251 @@
...
+
+            if cache is not None:
+                cache.set_synced_uuids(advert_contact_field_key, synced_uuids)
+
+    else:
+        assert len(uuids_to_sync) == 0
+        log.info("Found 0 uuids to sync in this run skipping...")
+
+
+def sync_advert_contacts_to_rapidpro(participants_by_column, uuid_table, pipeline_config, rapid_pro,
```
as2388, line 190: We call this `rapid_pro` rather than `rapidpro` everywhere else

as2388 (approved):
Thanks @IsaackMwenda, as you mentioned in a todo here we will need to think about how we can move this all into the db -> rapid pro sync in future.

PR #186, opened 2022-04-06
Support telegram group discussions

IsaackMwenda:
Initial implementation. Future tasks
- [ ] Dockerise
- [ ] Handle deleted messages
- [ ] Sync telegram contacts to rapidpro inorder to manually trigger demogs.
- [ ] Add Sync stats
- [ ] Dry run

Telethon Docs: https://docs.telethon.dev/en/stable/


configurations/test_telegram_group_pipeline_configuration.py
```diff
@@ -0,0 +1,113 @@
...
+        credentials_file_url="gs://avf-credentials/firebase-test.json",
+        table_name="_engagement_db_test",
+        uuid_prefix="avf-participant-uuid-"
+    ),
+    operations_dashboard=OperationsDashboardConfiguration(
+        credentials_file_url="gs://avf-credentials/avf-dashboards-firebase-adminsdk-gvecb-ef772e79b6.json",
+    ),
+    telegram_group_sources=[
+        TelegramGroupSource(
+            token_file_url="gs://avf-credentials/dev-telegram-tokens.json",
```
as2388, line 22:
```suggestion
            token_file_url="gs://avf-credentials/dev-telegram-token.json",
```
?

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,234 @@
...
+    :param group_id: Id of the telegram group to fetch messages from
+    :type group_id: str
+    :param dataset_end_date: Offset datetime, messages previous to this date will be retrieved. Exclusive
+    :type dataset_end_date: datetime | None
+    :param min_id: All the messages with a lower (older) ID or equal to this will be excluded.
+    :type min_id: int | None
+    :yields: Instances of telethon.tl.custom.message.Message
+    """
+    # Get group/channel entity
+    group_entity = await telegram.get_entity(PeerChannel(int(group_id)))
```
as2388, line 73: Can group id be an int in the first place?

configurations/test_telegram_group_pipeline_configuration.py
IsaackMwenda: This contains more than one credential (api_id,  api_hash, phone_number) so best to keep it as tokens

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,234 @@
...
+    # Ensure the client is authorized
+    if await telegram.is_user_authorized() == False:
+        await telegram.send_code_request(phone_number)
+        try:
+            await telegram.sign_in(phone_number, input(f"Enter the authorization code sent to "
+                                                              f"telegram a/c for {phone_number}: "))
+        # If the a/c has 2FA enabled sign_in() will raise a SessionPasswordNeededError.
+        # Input your telegram password to proceed.
+        except SessionPasswordNeededError:
+            await telegram.sign_in(password=input('Password: '))
```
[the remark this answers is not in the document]
IsaackMwenda: While authenticating the first time the client will send an auth code to the application phone number. The authentication info are stored in a session file to enable you authenticate automatically the next time the script runs. To re-authenticate you have to delete the session file. 

We can authenticate through a bot token but bots are restricted from accessing messages :(.

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,234 @@
...
+                    message_origin_details = {"message_id": telegram_message.id, "group_id":
+                                              telegram_message.peer_id.channel_id}
+                    message = _telegram_message_to_engagement_db_message(telegram_message, dataset.engagement_db_dataset,
+                                                                         uuid_table)
+                    _ensure_engagement_db_has_message(engagement_db, message, message_origin_details)
+
+                    # The api returns messages from newest to oldest, cache the id of the newest seen message for this search
+                    if cache is not None and cache_synced is False:
+                        cache.set_latest_group_message_id(group_cache_file_name, telegram_message.id)
+                        cache_synced = True
```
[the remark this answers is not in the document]
IsaackMwenda: Thanks - have updated this to only cache after adding all the available dataset messages in the group.

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,234 @@
...
+        # Input your telegram password to proceed.
+        except SessionPasswordNeededError:
+            await telegram.sign_in(password=input('Password: '))
+    log.info('Initialized telegram client...')
+
+    return telegram
+
+
+async def _fetch_message_from_group(telegram, group_id, dataset_end_date=None, min_id=None):
+    """
```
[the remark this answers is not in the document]
IsaackMwenda: Yup, unfortunately, the API only offers one date param, so we can't use start and end_date here. min_id helps us with the cache logic the next time we are pulling messages so that we don't fetch all messages again.

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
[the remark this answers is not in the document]
as2388: As discussed offline it would be great if there were a few credentials available, so that access to a DT phone number isn't a single point of failure. Can you add this as a todo for now?

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,234 @@
...
+        participant_uuid=participant_uuid,
+        text=telegram_message.message,
+        timestamp=telegram_message.date.isoformat(),
+        direction=MessageDirections.IN,
+        channel_operator=channel_operator,
+        status=MessageStatuses.LIVE,
+        dataset=dataset,
+        labels=[],
+        origin=MessageOrigin(
+            origin_id=f"message_id_{telegram_message.id}_timestamp_{telegram_message.date.isoformat()}",
```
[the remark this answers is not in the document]
as2388: I see, can you explain this in a comment because the fact the ids can be reused is not obvious!

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,240 @@
...
+def _is_avf_message(telegram_message):
+    """
+    :param telegram_message: A telegram message object to check if it was sent by an admin or is a channel broadcast
+                             Returns True if a message.from_id is None because admins send messages anonymously in the group
+                             and/or is of the type PeerChannel i.e channel broadcast messages.
+    :type telegram_message: telethon.tl.custom.message.Message
+    """
+    # Skip messages sent by AVF group admins / channel broadcasts
+    if (type(telegram_message.from_id) == PeerChannel or telegram_message.from_id is None):
+        return True
```
as2388, line 108:
```suggestion
    return (type(telegram_message.from_id) == PeerChannel or telegram_message.from_id is None):
```
so you can still get False back

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,240 @@
...
+        # If the a/c has 2FA enabled sign_in() will raise a SessionPasswordNeededError.
+        # Input your telegram password to proceed.
+        except SessionPasswordNeededError:
+            await telegram.sign_in(password=input('Password: '))
+    log.info('Initialized telegram client...')
+
+    return telegram
+
+
+async def _fetch_message_from_group(telegram, group_id, dataset_end_date=None, dataset_group_latest_seen_message_id=None):
```
as2388, line 63:
```suggestion
async def _fetch_message_from_group(telegram, group_id, dataset_end_date=None, dataset_group_start_message_id=None):
```
Suggesting this because when fetching I think all you care about is the data range, so using "latest_seen" in this case is a bit of a complexity this function doesn't need to worry about

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,240 @@
...
+
+        for group_id in dataset.search.group_ids:
+            group_cache_entry_name = f"{dataset.engagement_db_dataset}_{group_id}"
+            group_min_id = None if cache is None else cache.get_latest_group_message_id(group_cache_entry_name)
+
+            # Fetch group messages sent before the dataset_end_date and/or contain message.id greater than min_id in
+            # cache if available.
+            group_messages = await _fetch_message_from_group(telegram, group_id, dataset_end_date, group_min_id)
+
+            broad_cast_admin_messages = 0
```
as2388, line 217: broadcast is one word

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,240 @@
...
+            dataset_group_latest_seen_message_id = None
+            async for telegram_message in group_messages:
+                if _is_avf_message(telegram_message):
+                    broad_cast_admin_messages += 1
+                    continue
+
+                # Filter messages sent between this dataset start and end_time.
+                if dataset_start_date <= telegram_message.date < dataset_end_date:
```
as2388, line 225: Can you invert this to

```
if not <in-range>:
   continue

<mainline-logic>
```

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
```diff
@@ -0,0 +1,240 @@
...
+                    message_origin_details = {"message_id": telegram_message.id, "group_id":
+                                              telegram_message.peer_id.channel_id}
```
as2388, line 227: I think this should include the full message object, including its text and timestamp, not just the id

as2388 (approved):

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
IsaackMwenda: I disagree,  dataset_group_latest_seen_message_id best captures that this number is mutable with every run. dataset_group_start_message_id points to a predefined immutable id where each run will always start from.

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
IsaackMwenda: Storing as a string so that we can reuse existing cache string values.

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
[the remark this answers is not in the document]
IsaackMwenda: updated thanks.

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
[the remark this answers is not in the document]
IsaackMwenda: updated thanks.

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
as2388: Is it worth adding get_int and set_int to the cache? I'm slightly nervous if string/int conversions are going to cause us any problems, that's all.

DanielMwendwa (approved):

src/telegram_to_engagement_db/telegram_group_to_engagement_db.py
IsaackMwenda: Updated to int in configurations. (Sorry, I had misread this as int conversion for start_message_id)

PR #188, opened 2022-04-13
Fix not being able to run analysis without specifying a rapid pro target

as2388:


engagement_db_to_analysis.py
```diff
@@ -46,7 +46,11 @@
 
     uuid_table = pipeline_config.uuid_table.init_uuid_table_client(google_cloud_credentials_file_path)
     engagement_db = pipeline_config.engagement_database.init_engagement_db_client(google_cloud_credentials_file_path)
-    rapid_pro = pipeline_config.rapid_pro_target.rapid_pro.init_rapid_pro_client(google_cloud_credentials_file_path)
+
+    if pipeline_config.rapid_pro_target is None:
```
IsaackMwenda, line 50:
```suggestion
   rapid_pro = None if pipeline_config.rapid_pro_target is None else pipeline_config.rapid_pro_target.rapid_pro.init_rapid_pro_client(google_cloud_credentials_file_path)
```

IsaackMwenda (approved):

engagement_db_to_analysis.py
as2388: discussed offline, we'll keep this one as-is for readability.

PR #190, opened 2022-04-19
Dockerise telegram groups discussion to engagement db

IsaackMwenda:
Standard docker bash script slightly modified to create an interactive container to support user cmd input.

### DIFF
--- a/docker-run-telegram-group-to-engagement-db.sh
+++ b/docker-run-telegram-group-to-engagement-db.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+set -e
+
+PROJECT_NAME="$(<configurations/docker_image_project_name.txt)"
+IMAGE_NAME=$PROJECT_NAME-telegram-group-to-engagement-db
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        --incremental-cache-volume)
+            INCREMENTAL_ARG="--incremental-cache-path /cache"
+            INCREMENTAL_CACHE_VOLUME_NAME="$2"
+            shift 2;;
+        --)
+            shift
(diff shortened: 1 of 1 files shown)


docker-run-telegram-group-to-engagement-db.sh
```diff
@@ -0,0 +1,75 @@
...
+
+echo "Copying $CODE_SCHEMES_DIR -> $container_short_id:/app/code_schemes"
+docker cp "$CODE_SCHEMES_DIR" "$container:/app/code_schemes"
+
+echo "Copying $CONFIGURATION_FILE -> $container_short_id:/app/configuration.py"
+docker cp "$CONFIGURATION_FILE" "$container:/app/configuration.py"
+
+# Run the container
+echo "Starting container $container_short_id"
+docker start -a "$container"
```
as2388, line 65: The other scripts use -i too, do we not need that here?

as2388 (approved):

DanielMwendwa (approved):

docker-run-telegram-group-to-engagement-db.sh
IsaackMwenda: The interactivity does not seem to work on docker start, unfortunately - at least with the user input() function. Since we've created an interactive container then there is no need to pass it here.

as2388: Thanks, in that case I wonder if we ever needed the -i on the other scripts. In any case, no need to change anything here, this looks great

IsaackMwenda: Yes, we probably don't -  will test and send a pr.

PR #203, opened 2022-05-13
Handle message statuses correctly in common.get_messages_in_datasets

as2388:


src/common/get_messages_in_datasets.py
```diff
@@ -124,4 +127,31 @@ def get_messages_in_datasets(engagement_db, engagement_db_datasets, cache=None,
                                                                     f"'{msg.origin.origin_id}'"
             all_message_origins.add(msg.origin.origin_id)
 
+    # Filter out messages that don't meet the status conditions
+    for engagement_db_dataset, messages in engagement_db_messages_map.items():
+        # Filter out messages that do not have status "live" or "stale"
+        active_messages = [msg for msg in messages if msg.status in {MessageStatuses.LIVE, MessageStatuses.STALE}]
```
DanielMwendwa, line 133: The line where we append stale messages to live messages makes the code a little bit confusing. I think it's the active messages that should be appended to i.e, L143. I think we should introduce another variable name that is related to active messages, in this case, `actual_active_messages`.

I've added a few suggestions here and below to clearly articulate how I think this can be achieved.
```
active_messages = defaultdict(list)
active_messages["live_messages"].extend([msg for msg in messages if msg.status == "live"])
active_messages["stale_messages"].extend([msg for msg in messages if msg.status == "stale"])
```

src/common/get_messages_in_datasets.py
```diff
@@ -124,4 +127,31 @@ def get_messages_in_datasets(engagement_db, engagement_db_datasets, cache=None,
...
+        log.info(f"Filtered {engagement_db_dataset} for live/stale message snapshots: "
+                 f"{len(active_messages)}/{len(messages)} messages remain")
+
+        # Filter out stale messages for participants who sent a live message in that same dataset.
+        live_participants = {msg.participant_uuid for msg in messages if msg.status == MessageStatuses.LIVE}
+        live_messages = []
```
DanielMwendwa, line 139:
```
actual_active_messages = active_messages["live_messages"]
```

src/common/get_messages_in_datasets.py
```diff
@@ -124,4 +127,31 @@ def get_messages_in_datasets(engagement_db, engagement_db_datasets, cache=None,
...
+        stale_messages_count = 0
+        for msg in active_messages:
```
DanielMwendwa, line 141:
```
for msg in active_messages["stale_messages"]:
```

DanielMwendwa (approved):

src/common/get_messages_in_datasets.py
as2388: Thanks, I think this much clearer, so have reworked this along the lines suggested here

PR #212, opened 2022-06-10
Update docstrings

DanielMwendwa:


src/csv_to_engagement_db/csv_to_engagement_db.py
```diff
@@ -148,7 +148,7 @@ def _sync_csv_to_engagement_db(google_cloud_credentials_file_path, csv_source, e
     :type engagement_db: engagement_database.EngagementDatabase
     :param uuid_table: UUID table to use to re-identify the URNs so we can set the channel operator.
     :type uuid_table: id_infrastructure.firestore_uuid_table.FirestoreUuidTable
-    :param cache:
+    :param cache: csv sync cache
     :type cache: src.common.cache.Cache
```
as2388, line 152: Thanks for catching this - the cache can actually be None, here though, so this should be `:type cache: ...Cache | None` and the description should explain what happens if no cache is provided.

as2388 (approved):

src/csv_to_engagement_db/csv_to_engagement_db.py
DanielMwendwa: Thanks. Updated.

PR #253, opened 2023-01-06
Handle WhatsApp URNs in Rapid Pro -> Engagement DB Sync

IsaackMwenda:


as2388 (changes requested):
I think we also need to update the `channel_operator` in the created message to be "whatsapp" in cases where the message was sent to us over whatsapp

DanielMwendwa (approved):

PR #263, opened 2023-05-18
Simplify regression analysis data conversion

as2388:
Using IntVectors and StrVectors was unnecessarily complicated and will not work when we run multiple imputation. FactorVector is the correct conversion to use instead.

### DIFF
--- a/src/engagement_db_to_analysis/regression_analysis/r_utils.py
+++ b/src/engagement_db_to_analysis/regression_analysis/r_utils.py
@@ -1,18 +1,15 @@
 # TODO: Move to CoreDataModules once stable
 from collections import defaultdict
 
-from rpy2.robjects import DataFrame, IntVector, StrVector
+from rpy2.robjects import DataFrame, FactorVector
 
 
-def convert_dicts_to_r_data_frame(dicts, types):
+def convert_dicts_to_r_data_frame(dicts):
     """
     Converts a list of dictionaries to an R data-frame.
 
     :param dicts: Dictionaries to convert. Every dictionary must contain the same keys.
     :type dicts: dict of str -> (str | int | None)
-    :param types: Dictionary mapping each key in the `dicts` to the type of that key's values in the `dicts`, so we can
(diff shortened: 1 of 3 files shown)


src/engagement_db_to_analysis/regression_analysis/r_utils.py
```diff
@@ -30,16 +27,12 @@ def convert_dicts_to_r_data_frame(dicts, types):
...
 
     # 2. Convert each list of values to an R vector
     vectors = dict()
     for (dataset_name, values) in lists.items():
-        if types[dataset_name] == int:
-            vectors[dataset_name] = IntVector(values)
-        else:
-            assert types[dataset_name] == str, types[dataset_name]
-            vectors[dataset_name] = StrVector(values)
+        vectors[dataset_name] = FactorVector(values)
```
DanielMwendwa, line 35: Can you add a comment i.e, "Use FactorVector to create factors from categorical data" or 
"Factors are suitable for representing and analyzing categorical variables". it is very clear from the docstrings that `_get_participant_regression_data` returns categorical data. I also think `convert_dicts_to_r_data_frame` should display the same.

DanielMwendwa (approved):

src/engagement_db_to_analysis/regression_analysis/r_utils.py
as2388: I have updated the docstring for this function to add 'Each column in the returned data-frame contains a "FactorVector", which is the datatype used to represent categorical data in R.' 

I agree we should make it clear how the data will be interpreted during the conversion. I wonder if we should also rename the function itself to something like "convert_dicts_to_r_data_frame_of_factors" to make it clearer from the name? I can't think of a more succinct way to name it that makes it clear what you get back...

DanielMwendwa: sure `convert_dicts_to_r_data_frame_of_factors` is much clearer

PR #264, opened 2023-05-30
Add multiple imputation regression analysis

as2388:


src/engagement_db_to_analysis/regression_analysis/multiple_imputation_regression_analysis.py
```diff
@@ -0,0 +1,142 @@
...
+        participants, consent_withdrawn_field, rqa_analysis_config,
+        [d for d in demog_analysis_configs if d.dataset_name in demographic_datasets]
+    )
+
+    # Generate 20 copies of the input dataset, where each copy has had the missing data filled in with a different set
+    # of plausible values.
+    # Reset R's random number generator seed to ensure we get reproducible results.
+    log.info(f"Running multiple imputation for dataset '{rqa_analysis_config.dataset_name}'...")
+    base.set_seed(123)
+    multiple_imputed_data_frame = mice.mice(data_frame, m=20, printFlag=False)
```
DanielMwendwa, line 61: What is the basis for the choice of 20 for the number of imputations?  I think the optimal number of imputations can vary depending on the specific dataset and research question. Can we include it in our configs?

DanielMwendwa (approved):

